PR #1, opened 2016-09-22
Get SetOfAdminSubjects

millerts:
This is the initial endpoint for the assessment support service.  This service will support the cases when we need to go against the ItemBank schema.

Docker is missing and will be added in a separate pull request.  Going to work with @jeffjohnson9046 that tomorrow.


service/src/main/java/tds/assessment/repositories/AdminSubjectQueryRepository.java
```diff
@@ -0,0 +1,18 @@
...
+ * Queries for admin subject related entities
+ */
+public interface AdminSubjectQueryRepository {
+    /**
+     * Finds a {@link tds.assessment.SetOfAdminSubject} by key
+     *
+     * @param adminSubjectKey the unique key for the admin subject
+     * @return {@link tds.assessment.SetOfAdminSubject} if found otherwise empty
+     */
+    Optional<SetOfAdminSubject> findByKey(String adminSubjectKey);
```
jtreuting, line 17: Does this mean there was consensus on findXXX returns Optional while getXXX returns class directly?

service/src/main/java/tds/assessment/repositories/impl/AdminSubjectQueryRepositoryImpl.java
```diff
@@ -0,0 +1,56 @@
...
+
+        String SQL = "SELECT _key as `key`, \n" +
+            "testid as assessmentId, \n" +
+            "issegmented, \n" +
+            "selectionalgorithm as selectionAlgorithm \n" +
+            "FROM itembank.tblsetofadminsubjects \n" +
+            "WHERE _key = :key";
+
+
+        Optional<SetOfAdminSubject> setOfAdminSubjectOptional = Optional.empty();
```
jtreuting: I'm sure this was coded before the conversation about naming Optional variables.  Sounded like we were leaning towards maybeSetOfAdminSubjects?

service/src/main/java/tds/assessment/web/endpoints/AdminSubjectController.java
```diff
@@ -0,0 +1,46 @@
...
+import tds.assessment.SetOfAdminSubject;
+import tds.assessment.services.AdminSubjectService;
+import tds.assessment.web.resources.SetOfAdminSubjectResource;
+import tds.common.web.exceptions.NotFoundException;
+
+/**
+ * Controller handling admin subject related entities
+ */
+@RestController
+@RequestMapping("/assessments/admin-subject")
```
jtreuting, line 21: I like the hyphen for the URL instead of camelCase so just calling that out because I think this is the first time I've seen it.  Not sure if there were other routes using camelCase or not, but I think we should make the all lowercase hyphenated the standard like this.  @jeffjohnson9046

service/src/main/java/tds/assessment/web/endpoints/AdminSubjectController.java
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    /**
+     * Endpoint to find {@link tds.assessment.web.resources.SetOfAdminSubjectResource} by key
+     *
+     * @param key unique key for a {@link tds.assessment.SetOfAdminSubject}
+     * @return {@link org.springframework.http.ResponseEntity} containing a {@link tds.assessment.web.resources.SetOfAdminSubjectResource}
+     * @throws tds.common.web.exceptions.NotFoundException if entity cannot be found
+     */
+    @RequestMapping(value = "/{key}", method = RequestMethod.GET, produces = MediaType.APPLICATION_JSON_VALUE)
```
jtreuting, line 37: Based on one of your tests below the key appears to be the Test Key that comes from the test package and can really be anything.  I know we had an issue on the legacy app with a route ending in an email address because of the .com and thinking it was a static file resource.  Do we have to worry about that here?  I'm guessing not, but wanted to call it out.  Plus maybe we add a test for something like that as well.

service/src/test/java/tds/assessment/web/endpoints/AdminSubjectControllerTest.java
```diff
@@ -0,0 +1,62 @@
...
+    public void shouldFindAdminSubjectByKey() {
+        SetOfAdminSubject subject = new SetOfAdminSubject("theKey", "assessment", false, "alg");
+
+        when(service.findSetOfAdminObjectByKey("theKey")).thenReturn(Optional.of(subject));
+        ResponseEntity<SetOfAdminSubjectResource> response = controller.findSetOfAdminSubject("theKey");
+        verify(service).findSetOfAdminObjectByKey("theKey");
+
+        assertThat(response.getStatusCode()).isEqualTo(HttpStatus.OK);
+        assertThat(response.getBody().getSetOfAdminSubject().getKey()).isEqualTo("theKey");
+        assertThat(response.getBody().getId().getHref()).isEqualTo("http://localhost/assessments/admin-subject/theKey");
```
jtreuting, line 54: I believe in a previous test I looked at the href were localhost:8080.  Where is this set and should we standardize?

service/src/main/java/tds/assessment/repositories/AdminSubjectQueryRepository.java
jeffjohnson9046: I'm for it - I think it's a good convention.

service/src/main/java/tds/assessment/web/endpoints/AdminSubjectController.java
jeffjohnson9046: totally agree - I like the hyphen vs. camel case.

service/src/main/java/tds/assessment/repositories/AdminSubjectQueryRepository.java
millerts: Cool.  It was not a consensus until now.  I just added it because I agreed with the suggestion.

service/src/main/java/tds/assessment/web/endpoints/AdminSubjectController.java
millerts: First, the reason the email was an issue was because ART uses a mix of HTTP header and URL extension to determine accept type for the client.  So ".html" in that application would give you an HTML page whereas ".json" would give you json.

Secondly, if someone configures the DB with a client that contains an invalid URL character it will need to be encoded when the client calls this service.  I believe Spring will handle that for us, but I will verify when exam calls this application.

service/src/test/java/tds/assessment/web/endpoints/AdminSubjectControllerTest.java
millerts: Yes, the `localhost:8080` is what we use in Integration Tests because we are standing up an actual Web server.  However, this is just a unit test using a Spring provided `MockHttpServletRequest`.

jtreuting: Where is the emoji for "oops totally missed that"?

PR #2, opened 2016-09-29
Removed resource and add new integration test strategy

millerts:
Removed the HATEAOS stuff.  The bigger change was that I chose to use this pull request as what I think we should be doing as far as integration tests.


service/src/test/java/tds/assessment/repositories/impl/AdminSubjectQueryRepositoriesIntegrationTests.java
```diff
@@ -1,41 +0,0 @@
...
-
-import tds.assessment.SetOfAdminSubject;
-import tds.assessment.repositories.AdminSubjectQueryRepository;
-
-import static org.assertj.core.api.Assertions.assertThat;
-
-@RunWith(SpringJUnit4ClassRunner.class)
-@SpringBootTest
-@Transactional
-public class AdminSubjectQueryRepositoriesIntegrationTests {
```
millerts, line 20: I suggest that this integration test is not needed and the functionality is covered in the service integration below.

service/src/test/java/tds/assessment/services/impl/AdminSubjectServiceImplIntegrationTests.java
```diff
@@ -0,0 +1,108 @@
...
+    @Test
+    public void shouldFindAdminSubjectByKey() {
+        insertRecord();
+        Optional<SetOfAdminSubject> maybeSetOfAdminObject = adminSubjectService.findSetOfAdminByKey("(SBAC_PT)IRP-Perf-ELA-11-Summer-2015-2016");
+        assertThat(maybeSetOfAdminObject.get().getKey()).isEqualTo("(SBAC_PT)IRP-Perf-ELA-11-Summer-2015-2016");
+        assertThat(maybeSetOfAdminObject.get().getSelectionAlgorithm()).isEqualTo("virtual");
+        assertThat(maybeSetOfAdminObject.get().getAssessmentId()).isEqualTo("IRP-Perf-ELA-11");
+    }
+
+    private void insertRecord() {
```
millerts, line 54: Insert the record rather than putting inserts in the migration scripts.  Since this is the service layer we test the business logic and can leverage the @Transactional annotation

service/src/test/java/tds/assessment/web/endpoints/AdminSubjectControllerIntegrationTests.java
```diff
@@ -1,45 +1,54 @@
...
-@WebAppConfiguration
-@IntegrationTest("server.port:8080")
+import static org.hamcrest.CoreMatchers.is;
+import static org.mockito.Mockito.when;
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.get;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.jsonPath;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
+
+@RunWith(SpringRunner.class)
+@WebMvcTest(AdminSubjectController.class)
```
millerts, line 25: Looking through the Spring documentation I think this is the best way we can integration test controller logic.  The @WebMvcTest annotation makes Spring just stand up the Controller layer.  This would include the message converters, advice, and rest controllers.  Note that I use @MockBean and mockito to mock out the service.  

Main benefit here is that it is not hitting the database nor standing up the full Web container (so no port 8080 conflict).

millerts:
At this point I feel we only need two types of integration tests in our code base.  Service integration tests will test the service layer all the way down to the database.  Controller integration tests will simply test the controller mvc logic using mock services.

service/src/test/java/tds/assessment/repositories/impl/AdminSubjectQueryRepositoriesIntegrationTests.java
jtreuting: Since service methods can have conditional logic, there is a more likely case that we could miss a use-case and therefore miss testing a specific repository method that we would catch in writing repository integration test.  But I don't think this is a reason to not take this approach because I think you are right, this is double coverage really.  Just mentioning it because it makes the code coverage reports more valuable to make sure we don't miss areas.

jtreuting (approved):
this approach makes sense to me

service/src/test/java/tds/assessment/repositories/impl/AdminSubjectQueryRepositoriesIntegrationTests.java
millerts: True, I am hoping that the service integration tests would make sure to hit those conditionals.

PR #3, opened 2016-10-07
[TDS-352] GetInitialAbility - Adding startAbility value to model, upd…

burzum619:
…ating tests

### DIFF
--- a/service/src/main/java/tds/assessment/repositories/impl/AdminSubjectQueryRepositoryImpl.java
+++ b/service/src/main/java/tds/assessment/repositories/impl/AdminSubjectQueryRepositoryImpl.java
@@ -32,7 +32,8 @@ class AdminSubjectQueryRepositoryImpl implements AdminSubjectQueryRepository {
         String SQL = "SELECT _key as `key`, \n" +
             "testid as assessmentId, \n" +
             "issegmented, \n" +
-            "selectionalgorithm as selectionAlgorithm \n" +
+            "selectionalgorithm as selectionAlgorithm, \n" +
+            "startAbility \n" +
             "FROM itembank.tblsetofadminsubjects \n" +
             "WHERE _key = :key";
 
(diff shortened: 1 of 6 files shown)


jtreuting (approved):

service/src/main/java/tds/assessment/repositories/impl/AdminSubjectQueryRepositoryImpl.java
```diff
@@ -44,7 +45,8 @@ public AdminSubjectQueryRepositoryImpl(final DataSource dataSource) {
                     rs.getString("key"),
                     rs.getString("assessmentId"),
                     rs.getBoolean("issegmented"),
-                    rs.getString("selectionalgorithm")));
+                    rs.getString("selectionalgorithm"),
+                    rs.getFloat("startAbility")));
```
millerts, line 49: Do we want this to be null if the value in the DB is null?  If so, this won't work because `rs.getFloat()` returns 0 if the column value is null (docs [here](https://docs.oracle.com/javase/7/docs/api/java/sql/ResultSet.html#getFloat%28java.lang.String%29)).  If null is important I think we should leverage the solution mentioned [here](http://stackoverflow.com/questions/2920364/checking-for-a-null-int-value-from-a-java-resultset)

millerts: And if null isn't important I think we should make the `SetOfAdminSubject` property a primitive float instead of the Object Float.

millerts (changes requested):
One small clarification and potential change

service/src/main/java/tds/assessment/repositories/impl/AdminSubjectQueryRepositoryImpl.java
burzum619: Interesting point. Looking at production, stating, and deployment legacy itembank.tblsetofadminsubject tables, it appears that this value defaults to 0. However, taking a look at the CREATE statement for that table, it appears to be DEFAULT NULL. 

Looking a bit further into the itembank stored procedure that loads this table data (load_adminsubjects.sql in TDS_TestDeliverySystemDataAccess), it looks like the default value is set to 0. Based on this, I think we're safe to assume null isn't important and we can make the property a primitive float.

burzum619: startAbility changes to "float"

PR #4, opened 2016-10-13
Feature/tds docker

burzum619:
Docker related updates for the Assessment Service

**NOTE:** Currently configured to to use the "fwsbac" docker organization


service/src/main/docker/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM frolvlad/alpine-oraclejdk8:slim
+
+VOLUME /tmp
+ADD docker-startup.sh /
+ADD tds-assessment-service-0.0.1-SNAPSHOT.jar tds-assessment-service.jar
```
jtreuting, line 5: It seems like remembering to increment this jar name because of the version will become a headache, and error prone.  I believe we can have maven handle it during the build by using placeholders like ${project.version} in this file.  There is a way to tell maven to look at this file and process it.  A quick Google led me to this: http://stackoverflow.com/questions/3532135/using-maven-to-output-the-version-number-to-a-text-file, but this might only work if it's in the resources directory, not sure.

Anyway, not anything that we need to tackle now unless it's simple.

jtreuting (approved):

PR #5, opened 2016-10-20
Feature/get subject

millerts:
Open exam needs the subject name in the `tblsubject` table.  The query in the legacy app joins the `tbladminsubject` to the `tblsubject` table to get this data.  Instead of doing two hops decided to add a query param to signal the join


jeffjohnson9046 (approved):
These changes look fine to me, but I don't completely understand why sometimes we'd want to fetch the subject but other times we wouldn't.  I don't think the LEFT JOIN would be terribly expensive, so why not just always include the subject?

millerts:
Changed it to always do the join.

PR #6, opened 2016-10-26
Breaking SetOfAdminSubjects into Assessment, Segment classes

burzum619:
Repo has also been updated to map Segment rows in tblsetofadminsubjects to the segments collection in the Assessment.

Some uncertainties/doubts I have
- Currently, the Segments collection in each assessment is either empty (non-segmented test) or greater than 1. It is never 1, but I don't think this is particularly problematic.
- The endpoint was updated from /assessments/admin-subject/ to /assessments/assessment/. It seems slightly redundant, although it still makes sense to me as we are obtaining a single assessment (and not a segment, form, item, or other entity which we may add APIs for later)
- I kept the Assessment and Segments object immutable after creation (no setters). However, our constructors will become unwieldy very quickly as we add properties to these objects. Should we move to using Builders for these objects?


client/src/main/java/tds/assessment/Assessment.java
```diff
@@ -51,7 +61,7 @@ public String getAssessmentId() {
      * @return {@code true} if the assessment is segmented
      */
     public boolean isSegmented() {
-        return segmented;
+        return segments.size() > 1;
```
millerts, line 65: Suggest doing `return segments.isEmpty()` and add a check in the constructor that the List has to either be empty or contain more than 1 record.

client/src/main/java/tds/assessment/Assessment.java
```diff
@@ -1,40 +1,50 @@
...
+    public Assessment(String key,
+                      String assessmentId,
+                      List<Segment> segments,
+                      String selectionAlgorithm,
+                      float startAbility,
+                      String subjectName) {
         this.key = key;
         this.assessmentId = assessmentId;
-        this.segmented = segmented;
+        this.segments = segments;
```
millerts: What if we created a [umodifiableList](https://docs.oracle.com/javase/7/docs/api/java/util/Collections.html#unmodifiableList%28java.util.List%29) since the list should never change once created?

service/src/main/java/tds/assessment/repositories/impl/AssessmentQueryRepositoryImpl.java
```diff
@@ -0,0 +1,88 @@
...
+                "S.name, \n" +
+                "A.virtualtest AS assessmentKey\n" +
+            "FROM itembank.tblsetofadminsubjects A \n" +
+            "LEFT JOIN itembank.tblsubject S ON S._key = A._fk_Subject \n" +
+            "WHERE A.virtualtest = :key OR A._key = :key \n" +
+            "ORDER BY assessmentKey DESC";
+
+
+        List<Map<String,Object>> rows = jdbcTemplate.queryForList(SQL, parameters);
+        Assessment assessment = null;
```
millerts: Why declare the Assessment here?  I'd probably have it in the `else` block since it is is only used in it.  Same for the Segments List creation.

service/src/main/java/tds/assessment/web/endpoints/AssessmentController.java
```diff
@@ -9,36 +9,38 @@
...
 import tds.common.web.exceptions.NotFoundException;
 
 /**
  * Controller handling admin subject related entities
  */
 @RestController
-@RequestMapping("/assessments/admin-subject")
-class AdminSubjectController {
-    private final AdminSubjectService service;
+@RequestMapping("/assessments/assessment")
```
millerts: I think we should drop the `assessment`.  The only reason `admin-subject` was there is because I didn't understand that it represented an assessment.  Now I think the `key` is the id so `/assessments/key` makes sense to me.

service/src/test/java/tds/assessment/web/endpoints/AssessmentControllerIntegrationTests.java
```diff
@@ -0,0 +1,91 @@
...
+        Assessment assessment = new Assessment("(SBAC_PT)IRP-Perf-ELA-11-Summer-2015-2016", "IRP-Perf-ELA-11", segments, "virtual", 50F, "ELA");
+        when(assessmentSegmentService.findAssessmentByKey("(SBAC_PT)IRP-Perf-ELA-11-Summer-2015-2016")).thenReturn(Optional.of(assessment));
+
+        http.perform(get(new URI("/assessments/assessment/(SBAC_PT)IRP-Perf-ELA-11-Summer-2015-2016"))
+                .contentType(MediaType.APPLICATION_JSON))
+                .andExpect(status().isOk())
+                .andExpect(jsonPath("key", is("(SBAC_PT)IRP-Perf-ELA-11-Summer-2015-2016")))
+                .andExpect(jsonPath("assessmentId", is("IRP-Perf-ELA-11")))
+                .andExpect(jsonPath("subjectName", is("ELA")))
+                .andExpect(jsonPath("selectionAlgorithm", is("virtual")));
```
millerts: We should probably have some checks on the segment list.  You can see an example of json path assertions against a List [here](https://github.com/SmarterApp/TDS_ConfigService/blob/develop/service/src/test/java/tds/config/web/endpoints/ConfigControllerIntegrationTests.java#L151)

service/src/test/java/tds/assessment/web/endpoints/AssessmentControllerTest.java
```diff
@@ -10,50 +10,51 @@
...
 
     @Before
     public void setUp() {
         HttpServletRequest request = new MockHttpServletRequest();
         ServletRequestAttributes requestAttributes = new ServletRequestAttributes(request);
         RequestContextHolder.setRequestAttributes(requestAttributes);
 
-        service = mock(AdminSubjectService.class);
-        controller = new AdminSubjectController(service);
+        service = mock(AssessmentService.class);
```
millerts: Mind adding @RunWith(MockitoJunitRunner.class) and using the @Mock mockito annotation above the service?  @jeffjohnson9046 rolls with that style and I agreed to follow it.  This just hasn't been converted.

millerts (changes requested):
A couple small changes

millerts:
Since we already have the constructor I'd leave it for now, and the next lucky person that adds a property gets to add the builder.  Just isn't critical path imho

client/src/main/java/tds/assessment/Assessment.java
burzum619: Done, changed to !segments.isEmpty()

client/src/main/java/tds/assessment/Assessment.java
burzum619: done

service/src/test/java/tds/assessment/web/endpoints/AssessmentControllerTest.java
burzum619: done

service/src/main/java/tds/assessment/web/endpoints/AssessmentController.java
burzum619: I'm ok with this - but it may need to revisit this if we find a need to retrieve assessments based on something else (like a testid, or a segment key/id)

jtreuting:
You say "Currently, the Segments collection in each assessment is either empty (non-segmented test) or greater than 1."

Is there a reason for this?  I think in my mind we are representing the Assessment and not how it is curently stored in the DB.  So for me it feels more appropriate for a "non-segmented test" (meaning in reality it has 1 segment), to have 1 segment in the List.  Now I understand that in practice the assessmentKey and the assessmentSegmentKey will be the same in this case, but that really is because of convention (and how the loader restricts it if using the loader).

So I would argue that it should always have 1 or more segments and the check for if it is a segmented test is if the List contains more than 1 record.

@jeffjohnson9046, @millerts thoughts?

service/src/main/java/tds/assessment/web/endpoints/AssessmentController.java
millerts: I guess I would expect `/assessments/<unique key for assessment>`.  If you want to get an assessment by something else it would be `/assessments?segmentKey=key`.  Even though the segmentKey may be unique the assessment only has one unique identifier.  That or we would need to do `/assessment/<unique key>/<segmentKey>`

service/src/test/java/tds/assessment/web/endpoints/AssessmentControllerTest.java
burzum619: done

service/src/test/java/tds/assessment/web/endpoints/AssessmentControllerIntegrationTests.java
burzum619: done

service/src/main/java/tds/assessment/web/endpoints/AssessmentController.java
burzum619: Fair enough, I'll change to /assessments/

millerts:
@jtreuting , I don't feel strongly about it, but the structure you describe is what we discussed in the meeting.  So with what you're suggesting the properties on the `Assessment` object would match the properties of the single `segment` object on a non segmented assessment.

client/src/main/java/tds/assessment/Assessment.java
burzum619: Actually, kept this as size > 1 since we decided to have a non-segmented (single segment) assessment contain its own segment.

PR #7, opened 2016-11-02
Feature/fetch assessment languages

millerts:


client/src/main/java/tds/assessment/Property.java
```diff
@@ -0,0 +1,61 @@
+package tds.assessment;
+
+/**
+ * Represents active assessment properties
+ */
+public class Property {
```
millerts, line 6: Represents some of the data in `itembank.tblitemprops`

service/src/main/java/tds/assessment/repositories/impl/AssessmentMapper.java
```diff
@@ -0,0 +1,126 @@
...
+import java.util.stream.Collectors;
+
+import tds.assessment.Assessment;
+import tds.assessment.Property;
+import tds.assessment.Segment;
+
+/**
+ * Mapper used to map results from the AssessmentQuery for assessments by key
+ */
+class AssessmentMapper {
```
millerts, line 18: This was pretty complicated so I extracted the code that was in the repository into this separate class.  The mapping no longer requires a specific order to work properly.  The results from the query can have multiple queries for the same segment/assessment.  For example, if there is two different languages associated with the assessment there will be two `rows`.

service/src/main/java/tds/assessment/repositories/impl/AssessmentQueryRepositoryImpl.java
```diff
@@ -38,9 +38,13 @@ public AssessmentQueryRepositoryImpl(final DataSource dataSource) {
...
                 "startAbility, \n" +
                 "S.name AS subject, \n" +
-                "A.virtualtest AS assessmentKey\n" +
+                "A.virtualtest AS assessmentKey, \n" +
+                "P.propname, \n" +
+                "P.propvalue, \n" +
+                "P.propdescription \n" +
             "FROM itembank.tblsetofadminsubjects A \n" +
             "LEFT JOIN itembank.tblsubject S ON S._key = A._fk_Subject \n" +
+            "LEFT JOIN itembank.tblitemprops P ON P.isactive = 1 and propname = 'Language' AND P._fk_AdminSubject = A._key \n" +
```
millerts, line 47: The left join is for segmented assessments.  The base assessment doesn't have props but instead the individual segments do however we also want that base assessement record.

client/src/main/java/tds/assessment/Assessment.java
```diff
@@ -85,4 +87,54 @@ public float getStartAbility() {
     public String getSubjectName() {
```
burzum619: Minor quibble and nothing really to do with your changes - but thoughts on changing this property to just "subject"? I dont think theres any other data associated with a "subject" other than its name.

burzum619 (approved):

client/src/main/java/tds/assessment/Assessment.java
millerts: The reason I used subjectName was because there are other columns in that table like key, client, and loadconfig.  That being said we can just change the code later if we need more than name.

jeffjohnson9046 (approved):

PR #8, opened 2016-11-05
Adding Forms to AssessmentQueryRepository

burzum619:
Forms are now retrieved in the AssessmentQueryRepository.findByKey() method. List of forms are passed into the Mapper, which will assign the forms to the appropriate segments before returning the entire Assessment.

Multiple legacy table columns that seemingly aren't used have been excluded from the port. 

Examples of usage in legacy app:
ItemBankDaoImpl - line 109
StudentDLL - line 4054, 4065

Eventually, Form should have a collection of Items as well. Perhaps we can explore using Hibernate  or some other ORM framework.

### DIFF
--- a/service/src/main/java/tds/assessment/repositories/impl/AssessmentQueryRepositoryImpl.java
+++ b/service/src/main/java/tds/assessment/repositories/impl/AssessmentQueryRepositoryImpl.java
@@ -77,7 +77,7 @@ class AssessmentQueryRepositoryImpl implements AssessmentQueryRepository {
                         "    forms.updateconfig AS updateVersion, \n" +
                         "    forms.cohort \n" +
                         "FROM itembank.tblsetofadminsubjects segments \n" +
-                        "LEFT JOIN itembank.testform forms ON segments._key = forms._fk_adminsubject \n" +
+                        "JOIN itembank.testform forms ON segments._key = forms._fk_adminsubject \n" +
                         "WHERE segments.virtualtest = :key OR segments._key = :key";
 
 
--- a/service/src/main/java/tds/assessment/repositories/impl/AssessmentQueryRepositoryImpl.java
+++ b/service/src/main/java/tds/assessment/repositories/impl/AssessmentQueryRepositoryImpl.java
@@ -14,6 +14,7 @@ import java.util.Map;
 import java.util.Optional;
 
 import tds.assessment.Assessment;
+import tds.assessment.Form;
 import tds.assessment.repositories.AssessmentQueryRepository;
 
 @Repository
@@ -37,6 +38,9 @@ class AssessmentQueryRepositoryImpl implements AssessmentQueryRepository {
                 "testid AS assessmentSegmentId, \n" +
                 "selectionalgorithm AS selectionAlgorithm, \n" +
                 "startAbility, \n" +
+                "A.testposition AS segmentPosition, \n" +
+                "A.minItems, \n" +
+                "A.maxItems, \n" +
                 "S.name AS subject, \n" +
                 "A.virtualtest AS assessmentKey, \n" +
                 "P.propname, \n" +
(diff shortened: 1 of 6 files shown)


service/src/main/java/tds/assessment/repositories/impl/AssessmentQueryRepositoryImpl.java
```diff
@@ -55,9 +59,38 @@ public AssessmentQueryRepositoryImpl(final DataSource dataSource) {
...
+        final String formsSQL =
+                "SELECT \n" +
+                        "    segments._key AS segmentKey, \n" +
+                        "    forms._key AS `key`, \n" +
+                        "    forms.formid AS id, \n" +
+                        "    forms.language, \n" +
+                        "    forms.loadconfig AS loadVersion, \n" +
+                        "    forms.updateconfig AS updateVersion, \n" +
+                        "    forms.cohort \n" +
+                        "FROM itembank.tblsetofadminsubjects segments \n" +
```
millerts, line 79: Since we're specifically trying to get forms here I think this should go against itembank.testform and join on segments.  No left join.

`FROM itembank.testform forms JOIN itembank.tblsetofadminsubjects segements on segments._key = forms._fk_adminsubject`

I don't see why we'd want to get records back if the join isn't successful which the left will do.

service/src/test/java/tds/assessment/repositories/AssessmentQueryRepositoryImplIntegrationTests.java
```diff
@@ -92,9 +111,52 @@ public void shouldFindNonSegmentedAssessmentByKey() {
...
         assertThat(seg.getLanguages()).containsOnly(new Property("Language", "ENU"));
         assertThat(seg.getSelectionAlgorithm()).isEqualTo(maybeAssessment.get().getSelectionAlgorithm());
+
+        Form form1 = null;
+        Form form2 = null;
+        Form form3 = null;
+
+        for(Form form : seg.getForms()) {
+            switch (form.getKey()) {
+                case "187-510":
```
millerts: Is there a reason we need to check against three different form records?  Does the `form2` differ from `form1` enough that it is exercising something different in the query?

If this check is just making sure that all three form records are mapped properly I don't think we need anything other than `form1`.  We should insert and check the bare minimum otherwise we end up with a lot of noise in the tests.

service/src/test/java/tds/assessment/repositories/AssessmentQueryRepositoryImplIntegrationTests.java
```diff
@@ -134,9 +199,94 @@ public void shouldFindSegmentedAssessmentByKey() {
...
+        allSegments.addAll(formsSeg1);
+        allSegments.addAll(formsSeg2);
+        Form form1 = null;
+        Form form2 = null;
+        Form form3 = null;
+        Form form4 = null;
+        Form form5 = null;
+        Form form6 = null;
+
+        for(Form form : allSegments) {
```
millerts, line 196: Same question here.  Are the six forms here different resulting in different logic branches in the query being exercised?  Looking at the query the forms are a simple join so I don't know why we need all these different records.

millerts: Nothing wrong with any of this stuff. Just that the tests are more complex than I think they need to be which results in tests that are harder to understand and maintain.

millerts (changes requested):
Some questions about the complexity in our tests.  Also, the controller integration test should be modified to check the assessment json with the forms included.

That or as a team we should ditch those tests if people don't think they provide value.

service/src/main/java/tds/assessment/repositories/impl/AssessmentQueryRepositoryImpl.java
burzum619: You're absolutely right, done

PR #10, opened 2016-11-15
ItemConstraint/Property APIs, adding Item within Segments

burzum619:
- Segments now include the list of items (along with segment specific metadata from tblsetofadminitems)
- APIs for retrieving ItemConstraints and ItemProperties for use w/ Segment pool algorithm
- NOTE: item constraints table is in configs DB, but really is assessment specific.
- Renamed "Property" class to "ItemProperty" as it represents properties from tblitemprops.


client/src/main/java/tds/assessment/Item.java
```diff
@@ -0,0 +1,168 @@
...
+    private String segmentKey;
+    private String groupId;
+    private String groupKey;
+    private int position;
+    private boolean fieldTest;
+    private boolean active;
+    private boolean required;
+    private String strand;
+
+    private Item(Builder builder) {
```
millerts, line 23: We need a private empty constructor for json parsing.

client/src/main/java/tds/assessment/Segment.java
```diff
@@ -119,8 +128,9 @@ public int getMaxItems() {
         private int position;
         private int minItems;
         private int maxItems;
-        private List<Property> languages;
+        private List<ItemProperty> languages;
```
millerts, line 136: Should this be at this level anymore after your changes?  If not, then we should have a way to get properties by type.

service/src/main/java/tds/assessment/web/endpoints/ItemController.java
```diff
@@ -0,0 +1,74 @@
...
+
+    /**
+     * Endpoint to find the list of {@link tds.assessment.ItemConstraint}s for the assessment and client
+     *
+     * @param clientName    the client environment name
+     * @param assessmentId  the assessment id of the assessment that the constraints belong to
+     * @return {@link org.springframework.http.ResponseEntity} containing a list of  {@link tds.assessment.ItemConstraint}}
+     * @throws tds.common.web.exceptions.NotFoundException if no results are found
+     */
+    @RequestMapping(value = "/constraints/{clientName}/{assessmentId}", method = RequestMethod.GET, produces = MediaType.APPLICATION_JSON_VALUE)
```
millerts, line 40: I think the URL for this things should be `assessments/{assessmentid}/{client}/items/constraints` or `assessments/{client}/{assessmentid}/items/constraints`.

Also, shouldn't this be all contained in the assessment fetching the items?  I don't follow the lines that were drawn here with what we are fetching when we get an assessment vs this code.

service/src/main/java/tds/assessment/web/endpoints/ItemController.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    /**
+     * Endpoint to find the list of {@link tds.assessment.ItemProperty} objects for the {@link tds.assessment.Segment}
+     *
+     * @param segmentKey  the segment key of the segment that the constraints belong to
+     * @return {@link org.springframework.http.ResponseEntity} containing a list of  {@link tds.assessment.ItemProperty}}
+     * @throws tds.common.web.exceptions.NotFoundException if no results are found
+     */
+    @RequestMapping(value = "/properties/{segmentKey}", method = RequestMethod.GET, produces = MediaType.APPLICATION_JSON_VALUE)
```
millerts, line 61: Do we not get this with an assessment?  Also, shouldn't the URL be `assessments/{key}/segments/{segmentkey}/items/properties`?

millerts: Also, if we're going to have these endpoints why don't we just have one for items that gives you the items with all these values populated?

millerts (changes requested):

jtreuting:
I was going to review, but I think we are having a discussion about what to include with the Assessment and what not to, which might change things here.  So I'm going to wait for that first.

PR #13, opened 2016-11-17
Assessment Service refactoring

burzum619:
- Added Item, ItemProperties, ItemConstraints into the Assessment domain hierarchy
- Removed individual endpoints for ItemConstraint, ItemProperties. Moved integration/unit tests into ItemQueryRepositoryImplIntegrationTests
- Broke out Forms retrieval into its own repo.
- Full Assessment is now assembled in AssessmentService. Builders have been removed on Assessment, Segment, and Items and setters were added.

Concerns and quibbles:
- Should "clientname" be added as a param to the assessment GET API? The item constraints in particular are Clientname specific, and its theoretically possible for multiple assessments with the same assessmentId to be loaded into the system with different clientnames.
- Should ItemConstraint be possibly moved into AssessmentRepo (or its own repo even?) It's not tied into any particular items.
- ItemConstraints are currently retrieved by assessmentId (while other data is retrieved with assessmentKey). Not a big deal as the service calling findItemConstraints already has retrieved the Assessment (and assessmentId).


service/src/main/java/tds/assessment/services/impl/AssessmentServiceImpl.java
```diff
@@ -1,25 +1,78 @@
...
+        for (Segment segment : assessment.get().getSegments()) {
+            segment.setForms(forms.stream()
+                .filter(form -> form.getSegmentKey().equals(segment.getKey())).collect(Collectors.toList()));
+            segment.setItems(items.stream()
+                .filter(item -> item.getSegmentKey().equals(segment.getKey())).collect(Collectors.toList()));
+        }
+    }
+
+    private static void mapItems(List<Item> items, List<ItemProperty> itemProperties) {
+        Multimap<String, ItemProperty> itemIdToPropertyMap = ArrayListMultimap.create();
```
burzum619, line 69: In hindsight, instead of using a Multimap to map the properties to each item, we could add an "itemId" property to the ItemProperty to link it back to its parent Item. Then we can use lambdas to map properties to the Items as we do with Forms and Items in lines 59-62 of this same file.

client/src/main/java/tds/assessment/Assessment.java
```diff
@@ -12,9 +12,18 @@
     private String selectionAlgorithm;
     private float startAbility;
     private String subject;
+    private List<ItemConstraint> itemConstraints;
```
millerts: Can we set this up with an empty list?  Either in the getter or like we do with Segments?  Basically, I think we shouldn't return null.

millerts: I also wonder if we should make these arrays instead of lists.

client/src/main/java/tds/assessment/Segment.java
```diff
@@ -12,33 +12,23 @@
     private float startAbility;
     private String subject;
     private String assessmentKey;
-    private int position;
-    private int minItems;
+    private int position;private int minItems;
```
millerts: Please add a line break between properties

service/src/main/java/tds/assessment/repositories/impl/ItemQueryRepositoryImpl.java
```diff
@@ -0,0 +1,133 @@
...
+                rs.getString("description"),
+                rs.getString("itemId")
+            )
+        );
+    }
+
+    @Override
+    public List<ItemConstraint> findItemConstraintsForAssessment(final String assessmentId) {
+        Map<String, Object> parameters = new HashMap<>();
+//        parameters.put("clientName", clientName);
```
millerts: I assume this will be added back in a subsequent pull request?

service/src/main/java/tds/assessment/services/impl/AssessmentServiceImpl.java
```diff
@@ -1,25 +1,78 @@
...
+            maybeAssessment.get().setItemConstraints(itemConstraints);
+
+            mapItems(items, itemProperties);
+            mapAssessment(maybeAssessment, forms, items);
+        }
+
+        return maybeAssessment;
+    }
+
+    private static void mapAssessment(Optional<Assessment> assessment, List<Form> forms, List<Item> items) {
```
millerts: Can we rename this to something like `assignFormsAndItemsToAssessmentSegments()`?

service/src/main/java/tds/assessment/services/impl/AssessmentServiceImpl.java
```diff
@@ -1,25 +1,78 @@
...
+        this.itemQueryRepository = itemQueryRepository;
     }
 
     @Override
     public Optional<Assessment> findAssessmentByKey(String assessmentKey) {
-        return assessmentQueryRepository.findAssessmentByKey(assessmentKey);
+        Optional<Assessment> maybeAssessment = assessmentQueryRepository.findAssessmentByKey(assessmentKey);
+
+        if (maybeAssessment.isPresent()) {
+            List<Form> forms = formQueryRepository.findFormsForAssessment(assessmentKey);
```
millerts, line 45: As we enter the conditional can we do `Assessment assessment = maybeOptional.get();`?  I think it'd read better below.

service/src/main/java/tds/assessment/services/impl/AssessmentServiceImpl.java
millerts: Also, since this is private we should just pass the assessment not the Optional since this is only called if assessment isn't null.

service/src/main/java/tds/assessment/services/impl/AssessmentServiceImpl.java
millerts: With the way Java references work everything else will work the same

service/src/main/java/tds/assessment/services/impl/AssessmentServiceImpl.java
millerts: I'm not following this comment.  It looks like the property already has an itemId.

service/src/test/java/tds/assessment/repositories/ItemQueryRepositoryImplIntegrationTests.java
```diff
@@ -0,0 +1,247 @@
...
+        assertThat(item1.getGroupKey()).isEqualTo("GK-4");
+        assertThat(item1.getItemType()).isEqualTo("MC");
+        assertThat(item1.getPosition()).isEqualTo(1);
+        assertThat(item1.getSegmentKey()).isEqualTo("(SBAC_PT)IRP-Perf-ELA-11-Summer-2015-2016");
+        assertThat(item1.getStrand()).isEqualTo("silver strand");
+        assertThat(item1.isActive()).isTrue();
+        assertThat(item1.isFieldTest()).isTrue();
+        assertThat(item1.isRequired()).isTrue();
+    }
+
```
millerts, line 141: Can we remove some of these extra line breaks?

client/src/main/java/tds/assessment/Assessment.java
```diff
@@ -12,9 +12,18 @@
...
     private List<Segment> segments = new ArrayList<>();
 
-    private Assessment() {}
+    public Assessment() {}
```
millerts, line 18: If we leave this as is the public constructor can go away since Objects all have an empty constructor.

General question about all mutable objects in this project, do we want to have everything to have setters?  Or should the constructor instead take in the minimal amount of values that would make for a valid object.  I'd use the DB (non null fields) as a starting place for those minimal properties.

We can discuss more in tomorrow's parking lot.

service/src/main/java/tds/assessment/repositories/impl/ItemQueryRepositoryImpl.java
```diff
@@ -0,0 +1,133 @@
...
+                    item.setGroupKey(rs.getString("groupkey"));
+                    item.setPosition(rs.getInt("position"));
+                    item.setActive(rs.getBoolean("isactive"));
+                    item.setFieldTest(rs.getBoolean("isfieldtest"));
+                    item.setRequired(rs.getBoolean("isrequired"));
+                    item.setStrand(rs.getString("strandname"));
+                    return item;
+                }
+        );
+
```
millerts, line 67: Remove extra line breaks.

service/src/main/java/tds/assessment/repositories/impl/ItemQueryRepositoryImpl.java
```diff
@@ -0,0 +1,133 @@
...
+@Repository
+public class ItemQueryRepositoryImpl implements ItemQueryRepository {
+    private NamedParameterJdbcTemplate jdbcTemplate;
+
+    @Autowired
+    public ItemQueryRepositoryImpl(final DataSource dataSource) {
+        jdbcTemplate = new NamedParameterJdbcTemplate(dataSource);
+    }
+
+    public List<Item> findItemsForAssessment(final String assessmentKey) {
```
millerts, line 29: Did you consider left joining tblitemprops and getting the items and their properties all in a single call?  It'd be one less db call we'd have to do.  Obviously it will add a bit of complexity to the mapper.

millerts (changes requested):
Some general questions and small tweaks.

millerts:
To your general comment questions:
- Yes, it looks like clientname should be added to the path.
- It sounds like itemConstraints are at an assessment level not an item level so I would put it in Assessment related repository.
- I agree it isn't a big deal.

client/src/main/java/tds/assessment/Assessment.java
burzum619: Todd, I think that's a great point. There are some constructors (like segment or item) that take in their respective keys as constructor arguments - I think if it makes sense, we should add as many required fields to the constructor as possible. At what point do we decide there are too many arguments to include in the constructor, though?

Regarding setters - I'd be fine with removing the setters that aren't needed. Curious to hear other people's thoughts on this.

client/src/main/java/tds/assessment/Segment.java
burzum619: Woops, nice catch. Fixed.

service/src/main/java/tds/assessment/repositories/impl/ItemQueryRepositoryImpl.java
burzum619: I did briefly consider this, but not sure how much it would really save us in the long run considering the Assessment will likely be pretty heavily cached. I'm fine with either direction.

service/src/main/java/tds/assessment/repositories/impl/ItemQueryRepositoryImpl.java
burzum619: I would like to include it in this pull request. I wanted to get some feedback (as discussed at standup) on whether or not we should include the clientName in the GET Assessment API url params. I've been leaning towards "yes".

service/src/main/java/tds/assessment/services/impl/AssessmentServiceImpl.java
burzum619: Done

service/src/main/java/tds/assessment/services/impl/AssessmentServiceImpl.java
burzum619: Yes on both points, thanks.

service/src/main/java/tds/assessment/repositories/impl/ItemQueryRepositoryImpl.java
millerts: I thought we left with a pretty clear understanding of how it would work after standup.

`assessments/{clientName}/{assessmentKey}`

The reason I asked for it to not be included is because the URL change results in changes across other projects stopping this from being merged until that work is complete.  Branch off this branch if you want.  I see no reason it all needs to be in this pull.

It isn't as big of deal anymore now that people are no longer waiting for this work to complete.

client/src/main/java/tds/assessment/Assessment.java
burzum619: I agree with initializing all these as empty lists (to avoid NPEs). Shouldn't in happen in practice since the query() repo calls return an empty list by default, but still a good idea.

I'm not really convinced that arrays are necessary here. I guess they wouldn't hurt, but from my experience/research they don't really seem to have much benefit. When working with objects I prefer to use Lists. Lists have more flexibility, and I don't think we'd see any sort of performance gain by switching to arrays. My understanding is that the ArrayList implementation uses arrays internally anyway. With that said, I'm open and could definitely be convinced to use arrays instead.

PR #14, opened 2016-11-18
Creating spring config class to wire up JdbcTemplate beans

burzum619:
Production and test code should no longer require autowiring a datasource directly.


service/src/test/java/tds/assessment/configuration/data/DataSourceConfiguration.java
```diff
@@ -0,0 +1,38 @@
+package tds.assessment.configuration.data;
```
millerts: This should be in common and imported like the other common spring classes.

service/src/main/java/tds/assessment/AssessmentServiceApplication.java
```diff
@@ -1,9 +1,13 @@
 package tds.assessment;
 
 import org.springframework.boot.SpringApplication;
+import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.boot.autoconfigure.jdbc.DataSourceAutoConfiguration;
+import org.springframework.context.annotation.ComponentScan;
 
 @SpringBootApplication
+@EnableAutoConfiguration(exclude={DataSourceAutoConfiguration.class})
```
millerts: Why does this need to be excluded?

millerts (changes requested):

service/src/main/java/tds/assessment/AssessmentServiceApplication.java
burzum619: Replacing with Import

service/src/test/java/tds/assessment/configuration/data/DataSourceConfiguration.java
burzum619: Done

service/src/main/java/tds/assessment/AssessmentServiceApplication.java
```diff
@@ -2,8 +2,12 @@
 
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.context.annotation.Import;
+
+import tds.common.configuration.DataSourceConfiguration;
 
 @SpringBootApplication
+@Import(DataSourceConfiguration.class)
```
millerts: Instead of importing here please add it to the imports in ConfigServiceApplicationConfiguration.  Each support service now has one of these (named appropriately) that centralizes imports and any bean configurations that don't fit under a specific use.

millerts: Oops.  AssessmentApplicationConfiguration not config.

PR #15, opened 2016-11-22
Adding Strand collections to assessment and segments.

burzum619:
- Added Strand collections to Segment and Assessment classes. The strand data contains information for the adaptive algorithm and segment pool calculations
- Refactored mapping logic for Assessment into an "Assembler" helper class.


service/src/main/java/tds/assessment/services/impl/AssessmentAssembler.java
```diff
@@ -0,0 +1,69 @@
...
+    /**
+     * Maps data to the {@link tds.assessment.Assessment} and {@link tds.assessment.Segment}s
+     *
+     * @param assessment the assessment object to map
+     * @param forms     the forms to map to the segments in the assessment
+     * @param items     the items to map to the segments in the assessment
+     * @param constraints the constraints to map for the assessment
+     * @param strands   the strands to map to each segment and assessment
+     */
+    static void assignDataToAssessmentSegments(Assessment assessment,
```
burzum619: I renamed this "assignDataToAssessmentSegments" to make it more generic, as it now includes item constraints and strands. I did not want to break out each of these assignments into individual calls (assignFormsToX, assignItemsToX, etc) because I then need to iterate over each segment in the assessment multiple times per piece of data. I'm open to suggestions in naming or anything else though.

service/src/main/java/tds/assessment/repositories/StrandQueryRepository.java
```diff
@@ -0,0 +1,18 @@
...
+ * Repository for reading {@link tds.assessment.Strand} data
+ */
+public interface StrandQueryRepository {
+
+    /**
+     * Retrieves the list of {@link tds.assessment.Strand}s for the {@link tds.assessment.Assessment}
+     * @param assessmentKey the key of the {@link tds.assessment.Assessment}
+     * @return  the list of the strands for the assessment and its segments
+     */
+    Set<Strand> findStrands(String assessmentKey);
```
jtreuting, line 17: Is clientName needed here?

service/src/main/java/tds/assessment/repositories/impl/StrandQueryRepositoryImpl.java
```diff
@@ -0,0 +1,57 @@
...
+                "   segments.virtualtest = :key OR \n" +
+                "   segments._key = :key";
+
+        return new HashSet<>(jdbcTemplate.query(SQL, parameters, (rs, row) ->
+                new Strand.Builder()
+                        .withName(rs.getString("name"))
+                        .withKey(rs.getString("_key"))
+                        .withMinItems(rs.getInt("minitems"))
+                        .withMaxItems(rs.getInt("maxitems"))
+                        // calling getObject() and casting to Double because .getDouble() defaults to 0 if null
```
jtreuting: comment should say "casting to Float because .getFloat() ..."

service/src/main/java/tds/assessment/services/impl/AssessmentAssembler.java
```diff
@@ -0,0 +1,69 @@
...
+                                                      List<Form> forms,
+                                                      List<Item> items,
+                                                      List<ItemConstraint> constraints,
+                                                      Set<Strand> strands) {
+        assessment.setItemConstraints(constraints);
+        assessment.setStrands(strands.stream()
+                .filter(strand -> strand.getSegmentKey().equals(assessment.getKey())).collect(Collectors.toSet()));
```
jtreuting, line 46: When this is a single segmented assessment, since the key of the assessment and the segment are the same, it seems like this might assign the strands to the assessment here as well as the segment below.  Is that true and/or does that matter?

service/src/main/java/tds/assessment/services/impl/AssessmentAssembler.java
```diff
@@ -0,0 +1,69 @@
...
+import tds.assessment.Item;
+import tds.assessment.ItemConstraint;
+import tds.assessment.ItemProperty;
+import tds.assessment.Segment;
+import tds.assessment.Strand;
+
+/**
+ * A helper class for assembling the {@link tds.assessment.Assessment}
+ */
+class AssessmentAssembler {
```
jtreuting, line 20: I didn't see test coverage for the assembler

client/src/main/java/tds/assessment/Strand.java
```diff
@@ -0,0 +1,112 @@
...
+    private String name;
+    private String key;
+    private String segmentKey;
+    private int minItems;
+    private int maxItems;
+    private Float adaptiveCut;
+
+    private Strand() {}
+
+    public Strand (Builder builder) {
```
millerts: This shouldn't be public.

service/src/main/java/tds/assessment/services/impl/AssessmentAssembler.java
millerts: Why does the assessment and the segments have strands?  Are there strands that are at an assessment level and strands that are at a segment level?  My hunch is yes.

To Jeff's question, I don't know how this data will be used but I think it will matter if the assessment level strands are treated differently than segmented strands.

service/src/main/java/tds/assessment/services/impl/AssessmentServiceImpl.java
```diff
@@ -47,32 +48,13 @@ public AssessmentServiceImpl(AssessmentQueryRepository assessmentQueryRepository
             List<ItemProperty> itemProperties = itemQueryRepository.findActiveItemsProperties(assessmentKey);
             List<ItemConstraint> itemConstraints = assessmentQueryRepository.findItemConstraintsForAssessment(clientName,
                     assessment.getAssessmentId());
-            assessment.setItemConstraints(itemConstraints);
+            Set<Strand> strands = strandQueryRepository.findStrands(assessmentKey);
 
-            assignPropertiesToItems(items, itemProperties);
-            assignFormsAndItemsToAssessmentSegments(assessment, forms, items);
+            AssessmentAssembler.assignPropertiesToItems(items, itemProperties);
```
millerts: I would combine all these calls into a single call.  Pass in all the items then have it return the assessment.  I know you'll be updating it via reference but I think it'd read clearer if it was something like `assessment = assessmentAssembler.assemble(data...);`

service/src/test/java/tds/assessment/repositories/StrandQueryRepositoryImplIntegrationTests.java
```diff
@@ -0,0 +1,137 @@
...
+                strand2 = strand;
+            }
+        }
+
+        assertThat(strand1.getKey()).isEqualTo("Strand1-key");
+        assertThat(strand1.getMinItems()).isEqualTo(2);
+        assertThat(strand1.getMaxItems()).isEqualTo(7);
+        assertThat(strand1.getSegmentKey()).isEqualTo(assessmentKey);
+        assertThat(strand1.getAdaptiveCut()).isEqualTo(-39.234F);
+        assertThat(strand2.getKey()).isEqualTo("Strand2-key");
```
millerts, line 88: Can you add a line break between the assertions on strand 1 and strand 2?  Makes it easier to read.

service/src/test/java/tds/assessment/repositories/StrandQueryRepositoryImplIntegrationTests.java
```diff
@@ -0,0 +1,137 @@
...
+                strand5 = strand;
+            }
+        }
+
+        assertThat(strand3.getKey()).isEqualTo("Strand3-key");
+        assertThat(strand3.getMinItems()).isEqualTo(0);
+        assertThat(strand3.getMaxItems()).isEqualTo(2);
+        assertThat(strand3.getSegmentKey()).isEqualTo(assessmentKey);
+        assertThat(strand3.getAdaptiveCut()).isNull();
+        assertThat(strand4.getKey()).isEqualTo("Strand4-key");
```
millerts, line 121: Same comment here and really throughout to have line breaks between assertions of different objects.

service/src/main/java/tds/assessment/services/impl/AssessmentAssembler.java
millerts: I believe it is being covered by the service calls.  That being said I believe that is incorrect.  This is my fault as I originally was the one that didn't right a unit test around the original mapper code.

Since 90% of the Assessment related mapping logic is within this particular class I think this should have its own unit test which tests the methods.  Then the service unit test should become much simpler because it'll just verify that the repositories are called and the mapper is called.  The last part may be a little tricky since these are all static which means they can't be mocked.

service/src/test/java/tds/assessment/services/impl/AssessmentServiceImplTest.java
```diff
@@ -277,4 +292,139 @@ public void shouldReturnAssessmentWithItemProps() {
         assertThat(retCon2.getPropertyName()).isEqualTo(constraint2.getPropertyName());
         assertThat(retCon2.getPropertyValue()).isEqualTo(constraint2.getPropertyValue());
     }
+
+    @Test
+    public void shouldReturnAssessmentWithStrands() {
```
millerts: Is there ever an instance where an assessment will not have strands?  If not, why have the strand specific stuff in its own test?  I would expect the `shouldReturnAnAssessment` to also include the strand logic if that is the expected result of an Assessment when calling `findAssessment`

client/src/main/java/tds/assessment/Assessment.java
```diff
@@ -14,6 +16,7 @@
     private String subject;
     private List<ItemConstraint> itemConstraints = new ArrayList<>();
     private List<Segment> segments = new ArrayList<>();
+    private Set<Strand> strands = new HashSet<>();
```
millerts: I know I did this first but I'm wondering if instead of setting sets and lists here if instead we should tweak the getters.

```
Set<Strand> getStrands() {
  return strands != null ? strands : new HashSet<>();
}

**Review by millerts — CHANGES_REQUESTED** [review]

File: service/src/main/java/tds/assessment/services/impl/AssessmentAssembler.java [hunk 12]
```diff
@@ -0,0 +1,69 @@
...
+
+    /**
+     * Maps each {@link tds.assessment.ItemProperty} to its respective {@link tds.assessment.Item}
+     *
+     * @param items
+     * @param itemProperties
+     */
+    static void assignPropertiesToItems(List<Item> items, List<ItemProperty> itemProperties) {
+        Multimap<String, ItemProperty> itemIdToPropertyMap = ArrayListMultimap.create();
```
millerts: Why use a multimap rather than just have a `Map<String, List<ItemProperty>>`?  It doesn't look like we're using any of the specific features that a MultiMap would by us.

service/src/main/java/tds/assessment/services/impl/AssessmentAssembler.java
```diff
@@ -0,0 +1,69 @@
...
+        itemProperties.stream()
+                .filter(property -> property.getItemId() != null)
+                .forEach(property -> itemIdToPropertyMap.put(property.getItemId(), property));
+
+        for (Item item : items) {
+            item.setItemProperties(new ArrayList<>(itemIdToPropertyMap.get(item.getId())));
```
millerts: Added benefit to using a Map instead of MultiMap is then you wouldn't have to instantiate a new List.

service/src/main/java/tds/assessment/services/impl/AssessmentAssembler.java
```diff
@@ -0,0 +1,69 @@
+package tds.assessment.services.impl;
```
millerts, line 1: Why is this pulled out of the service?  I'm having trouble seeing the benefit since this is only ever used within the AssessmentService.

I understand we chatted about this yesterday but starting to question my comments after seeing it coded.

service/src/main/java/tds/assessment/services/impl/AssessmentAssembler.java
```diff
@@ -0,0 +1,69 @@
...
+        assessment.setItemConstraints(constraints);
+        assessment.setStrands(strands.stream()
+                .filter(strand -> strand.getSegmentKey().equals(assessment.getKey())).collect(Collectors.toSet()));
+
+        for (Segment segment : assessment.getSegments()) {
+            segment.setForms(forms.stream()
+                    .filter(form -> form.getSegmentKey().equals(segment.getKey())).collect(Collectors.toList()));
+            segment.setItems(items.stream()
+                    .filter(item -> item.getSegmentKey().equals(segment.getKey())).collect(Collectors.toList()));
+            segment.setStrands(strands.stream()
```
millerts, line 53: We are iterating through the strands n+1 times where n is the number of segments.  Rather than do that I think we should iterate over the strands at the beginning of the method and create a `Map<String,Set<Strand>>` with the key being the `strand.segmentKey`.  That would result in iterating over the strands a single time and then subsequent map lookups would be O(1).

service/src/main/java/tds/assessment/repositories/StrandQueryRepository.java
burzum619: Nope, it is not.

service/src/main/java/tds/assessment/services/impl/AssessmentAssembler.java
burzum619: @jtreuting  - you're right, both the Segment and Assessment will have the same strands. It is true, but I do not believe it matters. We could add code to check if it is a single segmented test and ignore adding the strands to the Assessment, but I don't believe this is really that necessary.

@millerts - That is correct.

service/src/main/java/tds/assessment/services/impl/AssessmentAssembler.java
burzum619: @millerts Yep, the reason why I didn't create a test class for the assembler is because that code was being covered by the Service. I do agree that it makes sense to port over the testing of this logic to a test class specific to the Assembler.

service/src/main/java/tds/assessment/services/impl/AssessmentServiceImpl.java
burzum619: I like this idea. I started doing something like this, but thought it may be less clear to return an Assessment as I'm updating by reference. I'll make the change though.

service/src/main/java/tds/assessment/services/impl/AssessmentAssembler.java
burzum619: Just for simplicity sake, really. Multimap just makes it easier/cleaner to insert multiple records with the same key.

service/src/main/java/tds/assessment/services/impl/AssessmentAssembler.java
burzum619: I anticipate we will be adding additional logic to the assembler class in the future, and having all this mapping logic in the service seems like it may get a bit cluttered.

service/src/main/java/tds/assessment/services/impl/AssessmentAssembler.java
burzum619: Agreed this would be more optimal. However, strands should never exceed more about 5 or 6 records in total.

service/src/main/java/tds/assessment/services/impl/AssessmentAssembler.java
burzum619: Using ArrayListMultimap instead of interface declaration to prevent the need for this. Thanks for the suggestion!

client/src/main/java/tds/assessment/Assessment.java
burzum619: done (and done for all other collections in Segment, Assessment, Item)

service/src/test/java/tds/assessment/services/impl/AssessmentServiceImplTest.java
burzum619: I could have combined all the different properties into one huge unit test method, but I thought it would be simpler to test the mapping of each piece of data separately.

burzum619: Moved this test into AssessmentAssemblerTest - testing the assembly of the entire assessment in one method now.

PR #16, opened 2016-11-23
Creating Algorithm enum

burzum619:
Creating Algorithm enum to replace "algorithmName" string properties in Assessments and segments.

### DIFF
--- a/client/src/main/java/tds/assessment/Algorithm.java
+++ b/client/src/main/java/tds/assessment/Algorithm.java
@@ -8,39 +8,39 @@ public enum Algorithm {
     VIRTUAL("virtual"),
     FIXED_FORM("fixedform");
 
-    private String algorithmName;
+    private String type;
 
-    Algorithm(String algorithmName) {
-        this.algorithmName = algorithmName;
+    Algorithm(String type) {
+        this.type = type;
     }
 
     /**
-     * @return the name of the algorithm 
+     * @return the name of the algorithm
      */
-    public String getAlgorithmName() {
-        return algorithmName;
+    public String getType() {
+        return type;
     }
 
     @Override
     public String toString() {
-        return this.getAlgorithmName();
+        return this.getType();
     }
 
     /**
      * Returns a {@link tds.assessment.Algorithm} enum for the algorithm name
      *
-     * @param algorithmName
+     * @param algorithmType Th
      * @return
      */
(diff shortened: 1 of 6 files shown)


client/src/main/java/tds/assessment/Algorithm.java
```diff
@@ -0,0 +1,46 @@
...
+    public static Algorithm fromAlgorithm(String algorithmName) {
+        if (algorithmName != null) {
+            for (Algorithm algorithm : values()) {
+                if (algorithmName.equalsIgnoreCase(algorithm.getAlgorithmName())) {
+                    return algorithm;
+                }
+            }
+        }
+        // No value found for string
+        throw new IllegalArgumentException();
```
millerts: We should always add messages to our Exceptions.  Please add one here; something like `"No Algorithm for name " + name`

client/src/main/java/tds/assessment/Algorithm.java
```diff
@@ -0,0 +1,46 @@
...
+    private String algorithmName;
+
+    Algorithm(String algorithmName) {
+        this.algorithmName = algorithmName;
+    }
+
+    /**
+     * @return the name of the algorithm 
+     */
+    public String getAlgorithmName() {
```
millerts: What if we call this `getValue()`?

client/src/main/java/tds/assessment/Algorithm.java
```diff
@@ -0,0 +1,46 @@
...
+        return this.getAlgorithmName();
+    }
+
+    /**
+     * Returns a {@link tds.assessment.Algorithm} enum for the algorithm name
+     *
+     * @param algorithmName
+     * @return
+     */
+    public static Algorithm fromAlgorithm(String algorithmName) {
```
millerts: Suggest using `fromValue` or `fromAlgorithmName` depending on the result of my other comment.

millerts (changes requested):

client/src/main/java/tds/assessment/Algorithm.java
```diff
@@ -0,0 +1,46 @@
...
+        if (algorithmName != null) {
```
millerts: If we want to check for null suggest flipping this conditional to be 

```{java}
if (algorithmName == null) throw new IllegalArgumentException("algorithmName is null")
```

PR #17, opened 2016-12-01
Feature/client test properties to assessment

millerts:
Couple things contained in this pull:

* Join to client_testproperties
* Cleaned up the RowMapper


service/src/main/java/tds/assessment/repositories/impl/AssessmentMapper.java
```diff
@@ -20,51 +21,24 @@
...
 
                 if (!segments.containsKey(key)) {
-                    segments.put(key, new SegmentInformationHolder());
-                }
-
-                SegmentInformationHolder holder = segments.get(key);
-                String segmentKey = (String) row.get("assessmentSegmentKey");
-
-                //If holder is null this is the first record for this segment
-                Segment segment = new Segment(segmentKey);
```
millerts, line 43: This block is mostly the same between the segmented assessment parsing and a non segmented assessment parsing so pull it into a private method below.

service/src/main/java/tds/assessment/repositories/impl/AssessmentMapper.java
```diff
@@ -75,63 +49,92 @@
...
-                    (String) assessmentRow.get("propname"),
-                    (String) assessmentRow.get("propvalue"),
-                    (String) assessmentRow.get("propdescription")
-                );
-
-                segment.setLanguages(Collections.singletonList(language));
-            }
-
+            Segment segment = new Segment(assessmentRow.getString("assessmentSegmentKey"));
+            segment.setAssessmentKey(assessmentRow.getString("assessmentSegmentKey"));
```
millerts, line 53: This is unneeded and will remove

service/src/main/java/tds/assessment/repositories/impl/AssessmentMapper.java
```diff
@@ -75,63 +49,92 @@
...
+            ItemProperty language = new ItemProperty(
+                propName,
+                row.getString("propvalue"),
+                row.getString("propdescription")
+            );
+            segment.getLanguages().add(language);
+        }
+    }
+
+    private class ResultRow {
```
millerts, line 107: I added this class because doing the casting when the value in the map is null will cause a NPE.  Fortunately, the data we have locally and in the deployment db happens to be not null (though the columns allow null).  This seemed like the cleanest way to fetch.

service/src/main/java/tds/assessment/repositories/impl/AssessmentMapper.java
```diff
@@ -75,63 +49,92 @@
...
+        assessment.setAbilityIntercept(assessmentRow.getFloat("abilityintercept"));
+        assessment.setAccommodationFamily(assessmentRow.getString("accommodationfamily"));
+        assessment.setMaxOpportunities(assessmentRow.getInt("maxopportunities"));
+
         assessment.setSegments(assessmentSegments);
 
         return Optional.of(assessment);
     }
 
-    private class SegmentInformationHolder {
```
millerts, line 119: This wasn't needed once we went away from immutability in segments.  Axed it.

client/src/main/java/tds/assessment/Segment.java
```diff
@@ -25,13 +25,12 @@
...
     private Set<Strand> strands;
 
-    /**
-     * Empty constructor for frameworks
-     */
-    private Segment() {}
-
     public Segment(String key) {
         this.key = key;
+        languages = new ArrayList<>();
```
millerts, line 30: If we have collections in the object I'd prefer them to be empty non null.  Makes it easier in calling code so you don't have to do `if(segment.getStrands != null) { for(segment.strands) }`

burzum619 (approved):
Changes look good to me

PR #18, opened 2016-12-01
Segment Properties

burzum619:
- Now joining configs.client_segmentproperties per Segment
- Removed join on tblitemprops (this info is still available for each item in the assessment)
- Added clientname filter to client_testproperties/segmentproperties portions of the query

### DIFF
--- a/service/src/main/java/tds/assessment/repositories/impl/AssessmentQueryRepositoryImpl.java
+++ b/service/src/main/java/tds/assessment/repositories/impl/AssessmentQueryRepositoryImpl.java
@@ -63,15 +63,16 @@ class AssessmentQueryRepositoryImpl implements AssessmentQueryRepository {
                 "SP.ftstartdate AS segFieldTestStartDate, \n" +
                 "SP.ftenddate AS segFieldTestEndDate \n" +
             "FROM itembank.tblsetofadminsubjects A \n " +
-            "JOIN configs.client_testproperties CT ON CT.testid = A.testid \n " +
-            "  OR CT.testid = (\n " +
+            "JOIN configs.client_testproperties CT \n" +
+            "   ON (CT.testid = A.testid AND CT.clientname = :clientName) \n " +
+            "   OR (CT.testid = (\n " +
             "    SELECT parentTsa.testid FROM itembank.tblsetofadminsubjects tsa \n " +
             "    JOIN itembank.tblsetofadminsubjects parentTsa ON tsa.virtualtest = parentTsa._key\n " +
             "    WHERE tsa._key = A._key \n " +
-            "  ) \n " +
-            "LEFT JOIN configs.client_segmentproperties SP ON SP.segmentid = A.testid \n" +
+            "  ) AND CT.clientname = :clientName) \n " +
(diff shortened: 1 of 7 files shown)


service/src/main/java/tds/assessment/repositories/impl/AssessmentQueryRepositoryImpl.java
```diff
@@ -54,25 +54,25 @@ public AssessmentQueryRepositoryImpl(final NamedParameterJdbcTemplate jdbcTempla
...
-            "    join itembank.tblsetofadminsubjects parentTsa on tsa.virtualtest = parentTsa._key\n " +
-            "    where tsa._key = A._key \n " +
-            "  ) \n " +
+            "JOIN configs.client_testproperties CT \n" +
+            "   ON (CT.testid = A.testid AND CT.clientname = :clientName) \n " +
+            "   OR (CT.testid = (\n " +
+            "    SELECT parentTsa.testid FROM itembank.tblsetofadminsubjects tsa \n " +
+            "    JOIN itembank.tblsetofadminsubjects parentTsa ON tsa.virtualtest = parentTsa._key\n " +
+            "    WHERE tsa._key = A._key \n " +
+            "  ) AND CT.clientname = :clientName) \n " +
```
millerts: This is a personal preference but I prefer this CT.clientName with the parameter to be in the `WHERE` rather than the join.  To me it is easier to follow what is being narrowed.  Same for the segment join below.

service/src/test/java/tds/assessment/repositories/AssessmentQueryRepositoryImplIntegrationTests.java
```diff
@@ -128,16 +136,20 @@ public void shouldFindNonSegmentedAssessmentByKey() {
         assertThat(seg.getMaxItems()).isEqualTo(4);
         assertThat(seg.getFieldTestMinItems()).isEqualTo(1);
         assertThat(seg.getFieldTestMaxItems()).isEqualTo(4);
-        assertThat(seg.getLanguages()).hasSize(1);
-        assertThat(seg.getLanguages()).containsOnly(new ItemProperty("Language", "ENU"));
         assertThat(seg.getSelectionAlgorithm()).isEqualTo(assessment.getSelectionAlgorithm());
     }
 
     @Test
+    public void shouldNotFindAssessmentWrongClient() {
```
millerts, line 143: Nice catch

millerts (approved):
I have a minor preference on the query structure for readability but will defer to others which style we want to follow.

service/src/main/java/tds/assessment/repositories/impl/AssessmentQueryRepositoryImpl.java
millerts: Even if we don't chagne this join can we put the `AND CT.clientname = :clientName on a separate line so it is clear that it is being done?

jtreuting: I'm fine with saying the best practices we should try to follow is to only include the joining keys in the `JOIN` statement and any filters in the `WHERE` clause.

burzum619: Sounds good to me. I'll update the query.

PR #20, opened 2016-12-14
Feature/single segment refactor

jeffjohnson9046:
* Refactor to remove `AdaptiveSegment` and `FixedFormSegment` subclasses
* Refactor `AssessmentMapper` and `AssessmentAssembler`
* Refactor unit test data creation to represent data as returned from database
* Additional unit test coverage


service/src/main/java/tds/assessment/services/impl/AssessmentServiceImpl.java
```diff
@@ -22,19 +25,17 @@
...
     @Autowired
     public AssessmentServiceImpl(AssessmentQueryRepository assessmentQueryRepository,
                                  ItemQueryRepository itemQueryRepository,
                                  FormQueryRepository formQueryRepository,
                                  StrandQueryRepository strandQueryRepository) {
-        this.formQueryRepository = formQueryRepository;
         this.assessmentQueryRepository = assessmentQueryRepository;
-        this.itemQueryRepository = itemQueryRepository;
         this.strandQueryRepository = strandQueryRepository;
+        this.assessmentAssembler = new AssessmentAssembler(itemQueryRepository, formQueryRepository);
```
jeffjohnson9046: I don't love this part.  It seems like a smell to me to pass in an `ItemQueryRepository` and `FormQueryRepository` implementation to the `AssessmentServiceImpl`, only to "pay it forward" to the `AssessmentAssembler`.  I thought about making the `AssessmentAssembler` its own legit service (complete w/interface), but that seemed a bit like overkill - how many `AssessmentAssembler` implementations are we likely to have?  So this is what I went with, but I'm certainly willing to make a change if someone has any thoughts.

The other part that "smells" to me is the `new`ing up of the `AssessmentAssembler` - the purist in me says if the `AssessmentService` implementation depends on it, then the `AssessmentAssembler` should be injected.

client/src/main/java/tds/assessment/Item.java
```diff
@@ -41,95 +41,94 @@ public boolean isRequired() {
...
+
     /**
-     * @return the
+     * @return the collection of item properties for the items associated with this Item
      */
     public List<ItemProperty> getItemProperties() {
         return itemProperties != null ? itemProperties : new ArrayList<ItemProperty>();
     }
 
-
```
millerts, line 100: We should discuss how we want to group these setters/getters.  Moving these around obscures the real changes in this class.

I'd prefer moving things around due to preference not technical issues to be done separately from feature requests for this reason.

client/src/main/java/tds/assessment/Segment.java
```diff
@@ -138,70 +157,112 @@ public Instant getFieldTestStartDate() {
...
 
-    public void setSelectionAlgorithm(Algorithm selectionAlgorithm) {
-        this.selectionAlgorithm = selectionAlgorithm;
-    }
+    /**
+     * @return All the {@link tds.assessment.Form}s associated with this segment.
+     * <p>
+     *     This getter/setter is only used by {@link tds.assessment.Segment}s that are configured for the "fixedform"
+     *     algorithm.  Segments configured for the "adaptive2" algorithm do not have forms, meaning this can return
+     *     null.
```
millerts: This javadoc is wrong.  It will return an empty list if not an adaptive algorithm.  Also, I think we should ditch the `adaptive2` and reference the enum since coders should only care about the enum not the value.

client/src/main/java/tds/assessment/Segment.java
```diff
@@ -138,70 +157,112 @@ public Instant getFieldTestStartDate() {
...
+     * </p>
+     *
+     * @param languageCode The student's specified language
+     * @return The {@link tds.assessment.Form} for the specified language
+     * @throws java.lang.IllegalArgumentException if the languageCode does not match any of the forms contained within
+     * this Segment
+     */
+    public Form getForm(String languageCode) {
```
millerts, line 210: I think this should return an Optional and not throw.  How is one to know which languages are acceptable?  That or this class should provide a `boolean isLanguageSupported()` method to allow a caller to check before trying to get the form for an unsupported language.

client/src/main/java/tds/assessment/Segment.java
```diff
@@ -138,70 +157,112 @@ public Instant getFieldTestStartDate() {
...
-    public void setForms(List<Form> forms) {
-        this.forms = forms;
-    }
+    /**
+     * @return the collection of {@link tds.assessment.Strand}s containing adaptive algorithm metadata for the
+     * {@link tds.assessment.Segment}
+     * <p>
+     *     This method is only used by {@link tds.assessment.Segment}s that are configured for the "adaptive2"
+     *     algorithm.  Segments configured for the "fixedform" algorithm do not have strands, meaning this will return
+     *     null.
```
millerts: This method never returns null and same comment about using the enum rather than the String values.

client/src/main/java/tds/assessment/Segment.java
```diff
@@ -138,70 +157,112 @@ public Instant getFieldTestStartDate() {
...
+    /**
+     * Get a collection of {@link tds.assessment.Item}s based on this Segment's selection algorithm.
+     *
+     * @param languageCode The language code
+     * @return A @{code List<Item>} collection for the student's language
+     */
+    public List<Item> getItems(String languageCode) {
+        // RULE: When the Segment's algorithm is set to "fixedform", the items should come from the Form that represents
+        // the student's selected language.  Otherwise, the items are returned directly (because they will have been
+        // collected by the computeSegmentPool method).
```
millerts: Where does the computeSegmentPool method live?  It doesn't seem to live in this class.  If it isn't in this class I guess you could reference the interface where this is done.  That or remove it.

service/src/main/java/tds/assessment/repositories/impl/AssessmentMapper.java
```diff
@@ -18,82 +18,128 @@
...
+     *     in the same table), this method builds a "default" segment from the same record that is used to build the
+     *     assessment.
+     * </p>
+     *
+     * @param row The {@link tds.assessment.repositories.impl.AssessmentMapper.ResultRow} representing the assessment
+     *            record returned from the query.
+     * @return An {@link tds.assessment.Assessment} with a default {@link tds.assessment.Segment}
+     */
+    private Optional<Assessment> getAssessmentWithDefaultSegment(ResultRow row) {
+        if (row == null) {
```
millerts: When will this ever be called with a null value?  I'd argue it isn't possible and since this is private this conditional should be removed.

service/src/main/java/tds/assessment/repositories/impl/AssessmentMapper.java
```diff
@@ -18,82 +18,128 @@
...
-        assessment.setAbilitySlope(assessmentRow.getFloat("abilityslope"));
-        assessment.setFieldTestStartDate(assessmentRow.getJodaInstantFromTimestamp("ftstartdate"));
-        assessment.setFieldTestEndDate(assessmentRow.getJodaInstantFromTimestamp("ftenddate"));
-        assessment.setAbilitySlope(assessmentRow.getFloat("abilityslope"));
-        assessment.setAbilityIntercept(assessmentRow.getFloat("abilityintercept"));
-        assessment.setAccommodationFamily(assessmentRow.getString("accommodationfamily"));
-        assessment.setMaxOpportunities(assessmentRow.getInt("maxopportunities"));
-        assessment.setInitialAbilityBySubject(assessmentRow.getBoolean("initialabilitybysubject"));
-        assessment.setSegments(assessmentSegments);
+        assessment.setKey(row.getString("assessmentSegmentKey"));
```
millerts, line 70: What changed between this block and the removed block?

service/src/main/java/tds/assessment/repositories/impl/AssessmentMapper.java
```diff
@@ -18,82 +18,128 @@
...
+            ? 1
+            : row.getInt("segmentPosition");
+
+        // Populate the segment instance with data.  Because we are building a Segment, the "assessmentSegmentKey"
+        // represents the key (unique identifier) for this Segment (and not the Assessment).
+        Segment segment = new Segment(row.getString("assessmentSegmentKey"),
+            Algorithm.fromType(row.getString("selectionAlgorithm")));
         segment.setAssessmentKey(assessmentKey);
-        segment.setSelectionAlgorithm(row.getAlgorithm("selectionalgorithm"));
+        segment.setSegmentId(row.getString("assessmentSegmentId"));
```
millerts, line 122: Just a general comment, I've noticed a lot of moving things around that logically do not change anything.  It makes it very hard to tell the actual changes in code.  For example, why was startAbility and setPosition moved?  I do not see any logical difference in their placement.

service/src/main/java/tds/assessment/repositories/impl/AssessmentQueryRepositoryImpl.java
```diff
@@ -43,46 +43,56 @@ public AssessmentQueryRepositoryImpl(final NamedParameterJdbcTemplate jdbcTempla
...
+            "   ON S._key = A._fk_Subject \n" +
             "WHERE \n" +
-            "   (A.virtualtest = :key OR A._key = :key) AND \n" +
-            "   (CT.clientname = :clientName OR SP.clientname = :clientName) \n" +
-            "ORDER BY assessmentKey DESC";
+            "   (A.virtualtest = :key OR A._key = :key) \n" +
+            "   AND (CT.clientname = :clientName OR SP.clientname = :clientName) \n" +
+            "ORDER BY  \n" +
+            "   assessmentKey DESC, \n" +
+            "   A.testPosition";
```
millerts, line 89: Spoke with @jeffjohnson9046 and if I followed this correctly `A.testPosition` is the only thing that changed in the above code outside of format changes.

service/src/main/java/tds/assessment/repositories/impl/AssessmentQueryRepositoryImpl.java
```diff
@@ -43,46 +43,56 @@ public AssessmentQueryRepositoryImpl(final NamedParameterJdbcTemplate jdbcTempla
...
-                "CT.accommodationfamily, \n" +
-                "CT.maxopportunities, \n" +
-                "CT.abilityslope, \n" +
-                "CT.abilityintercept, \n" +
-                "CT.initialabilitybysubject, \n" +
-                "SP.ftstartdate AS segFieldTestStartDate, \n" +
-                "SP.ftenddate AS segFieldTestEndDate \n" +
-            "FROM itembank.tblsetofadminsubjects A \n " +
-            "JOIN configs.client_testproperties CT \n" +
+            "   A._key AS assessmentSegmentKey, \n" +
```
millerts, line 46: This will mess up the formatting in IntelliJ.  I just had IntelliJ format the code (`cmd + option + l`) which will shift all these strings back to the right as they were before this change (except now they have the extra space).  This is because IntelliJ default formatting indents continuation code blocks the default 4 spaces.

Whatever we decide to do with this I vote for making it work easily with the IntelliJ default settings.

service/src/main/java/tds/assessment/repositories/impl/ItemQueryRepositoryImpl.java
```diff
@@ -22,31 +24,96 @@ public ItemQueryRepositoryImpl(final NamedParameterJdbcTemplate jdbcTemplate) {
...
+            "   ON segments._key = admin._fk_adminsubject \n" +
+            "WHERE \n" +
+            "   (segments.virtualtest = :key " +
+            "       OR segments._key = :key) \n" +
+            "   AND \n" +
+            "       formitems._fk_testform IN (:formKeys) \n" +
+            "   AND \n" +
+            "       formitems.isactive = 1 \n" +
+            "   AND \n" +
+            "       admin.isactive = 1";
```
jeffjohnson9046: Wanted to call this out too.  It'd certainly be possible to change the `AND` condition to `formitems.isactive = admin.isactive`.  However, because of the way the database is set up, the `isactive` on the form items might be set to something different than the `isactive` on the admin items.  I'm not sure which one trumps the other.  To err on the side of caution, I decided to check the `isactive` for each table rather than make sure they were equal.

service/src/main/java/tds/assessment/repositories/impl/AssessmentQueryRepositoryImpl.java
jeffjohnson9046, line 89: This is the only part of the query that _actually_ changed; everything else is formatting.

service/src/main/java/tds/assessment/repositories/impl/ItemQueryRepositoryImpl.java
```diff
@@ -22,31 +24,96 @@ public ItemQueryRepositoryImpl(final NamedParameterJdbcTemplate jdbcTemplate) {
...
-                        "    itembank.tblitem I \n" +
-                        "    ON I._key = A._fk_item \n" +
-                        "JOIN itembank.tblsetofadminsubjects segments \n" +
-                        "    ON segments._key = A._fk_adminsubject \n" +
-                        "WHERE \n" +
-                        "   (segments.virtualtest = :key OR segments._key = :key) AND \n" +
-                        "   A.isactive = 1";
-
-        List<Item> items = jdbcTemplate.query(itemsSQL, parameters, (RowMapper<Item>) (rs, row) -> {
+            "SELECT \n" +
```
jeffjohnson9046, line 31: no actual code change on this query; formatting for consistency.

service/src/main/java/tds/assessment/services/impl/AssessmentAssembler.java
```diff
@@ -3,57 +3,159 @@
...
+    }
+
+    /**
+     * Build an {@link tds.assessment.Segment} that uses the "adaptive2" selection algorithm for serving up items.
+     *
+     * @param segment The assessment object to map
+     * @param strands The collection of strands that should be associated to the assessment and its constituent segments
+     * @param itemProperties The properties that are associated to each item
+     */
+    private void assembleAdaptiveSegment(Segment segment, List<ItemProperty> itemProperties, Set<Strand> strands) {
```
millerts: I would prefer this to be named `updateAdaptiveSegment()` which clearly implies that the segment reference is being updated.

service/src/main/java/tds/assessment/services/impl/AssessmentAssembler.java
```diff
@@ -3,57 +3,159 @@
...
-                .forEach(property -> itemIdToPropertyMap.put(property.getItemId(), property));
+    void assemble(Assessment assessment,
+                  Set<Strand> strands,
+                  List<ItemConstraint> itemConstraints) {
 
-        for (Item item : items) {
-            item.setItemProperties(itemIdToPropertyMap.get(item.getId()));
+        List<ItemProperty> itemProperties = itemQueryRepository.findActiveItemsProperties(assessment.getKey());
+
+        addAssessmentContent(assessment, itemConstraints, strands, itemProperties);
```
millerts: I would prefer this renamed to `updateAssessmentContent`.  It makes it clear that the method alters the assessment reference.

service/src/main/java/tds/assessment/services/impl/AssessmentAssembler.java
```diff
@@ -3,57 +3,159 @@
...
  */
 class AssessmentAssembler {
+    private final ItemQueryRepository itemQueryRepository;
+    private final FormQueryRepository formQueryRepository;
+    private ArrayListMultimap<String, ItemProperty> itemIdToPropertyMap;
+    private List<Item> adaptiveAssessmentItems;
+    private Map<String, List<Item>> fixedFormAssessmentItems;
+    private List<Form> fixedFormAssessmentForms;
+
+    AssessmentAssembler(ItemQueryRepository itemRepository, FormQueryRepository formRepository) {
```
millerts: I think all repository calls should be in the service and this should simply take all the results and compose the object.  One reason is that we're going to tie transaction boundaries to our services and secondly this wouldn't need to hold state.

I don't see why this needs to hold any state at an instance level.

service/src/main/java/tds/assessment/services/impl/AssessmentAssembler.java
```diff
@@ -3,57 +3,159 @@
...
+     * @param constraints The collection of {@link tds.assessment.ItemConstraint}s
+     * @param strands The collection of {@link tds.assessment.Strand}s that are associated to the assessment
+     * @param itemProperties The properties that are associated to each item
+     */
+    private void addAssessmentContent(Assessment assessment,
+                                      List<ItemConstraint> constraints,
+                                      Set<Strand> strands,
+                                      List<ItemProperty> itemProperties) {
         assessment.setItemConstraints(constraints);
         assessment.setStrands(strands.stream()
```
millerts, line 37: So strands can be at an assessment level and a segment level?

service/src/main/java/tds/assessment/services/impl/AssessmentAssembler.java
```diff
@@ -3,57 +3,159 @@
...
+     *
+     * @param items The items that need to be assigned item properties
+     * @param itemProperties The item properties that need to be mapped to their items
+     */
+    private void addItemPropertiesToItems(List<Item> items, List<ItemProperty> itemProperties) {
+        if (itemIdToPropertyMap == null) {
+            itemIdToPropertyMap = ArrayListMultimap.create();
+            itemProperties.stream()
+                .filter(property -> property.getItemId() != null)
+                .forEach(property -> itemIdToPropertyMap.put(property.getItemId(), property));
```
millerts: I would need to play with it but it feels like a collect using collectors [groupingBy](https://docs.oracle.com/javase/8/docs/api/java/util/stream/Collectors.html#groupingBy-java.util.function.Function-java.util.function.Supplier-java.util.stream.Collector- ) could solve this without having to mutate the outer map in a stream.  Added benefit is we could remove the ArrayListMultiMap

service/src/main/java/tds/assessment/services/impl/AssessmentServiceImpl.java
millerts: I feel this service should fetch all the data then pass all that data into the assembler.  Maybe I missed something but it didn't seem like the assembler was fetching anything per a piece of the assessment.

service/src/test/java/tds/assessment/builders/AssessmentRecordBuilder.java
```diff
@@ -0,0 +1,210 @@
...
+import java.util.Map;
+
+import tds.assessment.Algorithm;
+
+/**
+ * Build data structures that represent what Assessment data looks like in the database.  This class is intended tu
+ * support testing the {@code AssessmentMapper}.
+ */
+public class AssessmentRecordBuilder {
+    private static Timestamp segmentFieldTestStartDate;
```
millerts: Why are these all static?  That means that once set the values are tied to the class itself.  So if one test modifies the value of `segmentFieldTestStartDate` it could impact another test.

millerts (changes requested):
The formatting, method placement, and value assignment placement changes made in this pull request made it extremely difficult for me to see the logic changes.  Unless there is a bug or a necessity due to refactor I would prefer those type of changes be made outside of any feature request.  I tried to review the queries but I can't say I followed all of the changes outside of the one we discussed offline mostly due to all the format changes.

It is also obvious to me we as a team need to discuss formatting changes so this type of wide spread format and rearranging doesn't need to occur in the future.

client/src/main/java/tds/assessment/Assessment.java
```diff
@@ -215,4 +216,19 @@ public boolean isInitialAbilityBySubject() {
...
+     * <p>
+     *     The languages that are available to an assessment are determined by collecting the unique "language"
+     *     {@link tds.assessment.ItemProperty} of the {@link tds.assessment.Item}.
+     * </p>
+     */
+    public Set<String> getLanguages() {
+        return languages;
+    }
+
+    public void setLanguages(Set<String> languages) {
```
jtreuting: Is this the "language code" or the "language"?  I remember @burzum619 recommending a naming convention that was language code when we were using ENU or ESN.  If we are following that practice, does this need to chagne to `getLanguageCodes` and `setLanguageCodes`?

Update: As I got further down, I see that `languageCode` is used as the parameter to `getForm()`, so I'm assuming this is the languageCode and should be renamed.

client/src/main/java/tds/assessment/Form.java
```diff
@@ -125,4 +137,27 @@ public Long getUpdateVersion() {
...
+    }
+
+    /**
+     * Determine the number of {@link tds.assessment.Item}s associated with this form.
+     *
+     * @return The number of {@link tds.assessment.Item}s in this form.  If the list is null, return 0.
+     */
+    public int getLength() {
+        if (items == null) {
+            items = new ArrayList<>();
```
jtreuting: I'm not sure I see a benefit to setting this here instead of just returning 0.  `return item == null ? 0 : items.size()`

client/src/main/java/tds/assessment/Segment.java
```diff
@@ -24,12 +24,17 @@
...
     private Instant fieldTestStartDate;
     private Instant fieldTestEndDate;
+
+    // Fixed-form specific fields
     private List<Form> forms;
-    private List<Item> items;
+
+    // Adaptive-specific fields
     private Set<Strand> strands;
+    private List<Item> items;
```
jtreuting, line 33: Small thing, but I read this as saying that the Items are only specific to the Adaptive algorithm.  Maybe moving Items above or just putting an extra line break.

service/src/main/java/tds/assessment/repositories/impl/AssessmentMapper.java
jtreuting: @millerts, looks like `assessmentRow` was changed to `row` and then the `segments` are created below using the `getSegment` method.

service/src/main/java/tds/assessment/repositories/impl/ItemQueryRepositoryImpl.java
```diff
@@ -75,10 +140,13 @@ public ItemQueryRepositoryImpl(final NamedParameterJdbcTemplate jdbcTemplate) {
...
-            "JOIN itembank.tblsetofadminsubjects segments \n" +
-            "    ON segments._key = P._fk_adminsubject \n" +
+            "JOIN " +
+            "   itembank.tblsetofadminsubjects segments \n" +
+            "   ON segments._key = P._fk_adminsubject \n" +
             "WHERE \n" +
-            "   segments.virtualtest = :key OR segments._key = :key AND" +
+            "   segments.virtualtest = :key \n" +
+            "   OR segments._key = :key \n" +
+            "AND\n" +
```
jtreuting, line 106: Currently without an parenthesis this will get interpreted as 
    `segments.virtualtest = :key OR (segments._key = :key AND isActive = 1)` 
while I think the actual intent is 
    `(segments.virtualtest = :key OR segment._key = :key) AND isActive = 1`

Also, can you add the alias to the `isActive` since I can't tell which table that is from by just looking at this query.

service/src/test/java/tds/assessment/builders/AssessmentRecordBuilder.java
```diff
@@ -0,0 +1,210 @@
...
+import java.sql.Timestamp;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import tds.assessment.Algorithm;
+
+/**
+ * Build data structures that represent what Assessment data looks like in the database.  This class is intended tu
```
jtreuting: gnit pick: typo "tu"

jtreuting (changes requested):

service/src/main/java/tds/assessment/repositories/impl/ItemQueryRepositoryImpl.java
millerts: Does this query exist in the legacy application?  If so, what does it do?  I see no issue with the change unless the legacy application makes sure they're the same.  Even if it doesn't necessarily make sense to us that small tweak could introduce a subtle bug.

service/src/main/java/tds/assessment/services/impl/AssessmentServiceImpl.java
burzum619: Agreed with Tom. I think the idea is that AssessmentService should fetch all data (using the various repositories), and the Assembler is responsible for constructing the actual Assessment.

service/src/main/java/tds/assessment/repositories/impl/ItemQueryRepositoryImpl.java
burzum619: In general we've been filtering by "isActive = 1" wherever possible/appropriate. In legacy code, I've seen cases where "isActive=1" is filtered directly in the data retrieval query, and other places where that filtration is done when inserting into temporary tables. This looks good to me, Jeff.

client/src/main/java/tds/assessment/Segment.java
burzum619: @millerts The languageCodes are part of the parent Assessment, presumably one of those codes will be used here.

service/src/main/java/tds/assessment/services/impl/AssessmentAssembler.java
burzum619: @millerts That's correct.

service/src/main/java/tds/assessment/repositories/impl/ItemQueryRepositoryImpl.java
millerts: Good point.  I forgot about the usage.

client/src/main/java/tds/assessment/Segment.java
millerts: I understand.  I just don't think we need to throw an Exception and instead can return an Optional.empty() and allow the caller to handle it. 

I realize it doesn't matter since this is all internal.  I guess I prefer to not throw Exceptions on getters that can simply return empty.

service/src/main/java/tds/assessment/services/impl/AssessmentAssembler.java
burzum619: I agree with Todd's comments on the assembler containing repository calls. I don't think AssessmentAssembler should contain anything other that static methods as it should not hold state.

burzum619 (changes requested):
Great job Jeff. I think my biggest issue is with the repo calls in the AssessmentAssembler, but the rest of the requested changes are fairly minor/trivial.

client/src/main/java/tds/assessment/Segment.java
burzum619: Unfortunately I don't believe it can return Optional as we are restricted to Java 7 for code in the client jars. But I agree with returning empty lists or null and letting the callers handle it.

jeffjohnson9046: Unfortunately since this class is in the `client` library, returning an `Optional` will be a bit of a pain.  We could:

* Add guava (or some equivalent) library to get an `Optional` and some functional stuff (that may or may not be useful)
* Return `null`
* Return a new, empty `Form`

Also, isn't it exceptional to ask for a form with a language code the `Segment` doesn't know about?

service/src/main/java/tds/assessment/repositories/impl/AssessmentMapper.java
jeffjohnson9046: These were moved around as part of the refactoring/debugging effort.  I was trying to diagnose an issue w/the column labels.  It was easiest for me to verify that everything was spelled correctly and being pulled in properly when the `segment.set[whatever]` were put in the same order as the columns returned by the query.

service/src/main/java/tds/assessment/repositories/impl/AssessmentQueryRepositoryImpl.java
jeffjohnson9046: Correct - the `A.testPosition` is the only actual logic change to this query.

client/src/main/java/tds/assessment/Segment.java
jeffjohnson9046: Pursuant to our conversation that @burzum619 is going to change this method to handle the updated information we received about form selection from SBAC, I'm going to leave this the way it is for now.

PR #24, opened 2017-01-09
TDS-354: Add item/stimulus paths and isPrintable to Item

jeffjohnson9046:
groundwork for get page content in TDS_ExamService


service/src/main/java/tds/assessment/repositories/impl/ItemQueryRepositoryImpl.java
```diff
@@ -47,9 +50,23 @@ public ItemQueryRepositoryImpl(final NamedParameterJdbcTemplate jdbcTemplate) {
                 "JOIN \n" +
                 "   itembank.testformitem FI \n" +
                 "   ON FI._fk_item = I._key \n" +
-                "JOIN " +
+                "JOIN \n" +
                 "   itembank.tblsetofadminsubjects segments \n" +
                 "   ON segments._key = A._fk_adminsubject \n" +
+                "JOIN \n" +
+                "   itembank.tblclient C \n" +
```
millerts: I know you're following the pattern from before but I think with the size of these queries that it may be beneficial to name these aliases rather than using single letters.  It still can be short.

- A -> adminItems
- I -> item
- FI -> formItem
- C -> client
- B -> bank
- S -> stimulus

client/src/main/java/tds/assessment/Item.java
```diff
@@ -137,10 +141,47 @@ public void setItemProperties(List<ItemProperty> itemProperties) {
...
+
+    /**
+     * @return The full path and file name of the item's stimulus file (if there is one)
+     */
+    public String getStimulusFilePath() {
+        return stimulusFilePath;
+    }
+
+    public void setStimulusFilePath(String stimulusFilePath) {
+        if (stimulusFilePath == null) {
```
millerts, line 166: Why does this get null treatment where itemFilePath does not?

jeffjohnson9046: Technically, the `tblitem.filepath` column is nullable, so it probably should get the same treatment.  As it turns out, all items in `itembank.tbltem` have a `filepath` value.  Same for `itembank.tblstimulus`.  However, there are Items that do not have a "companion" stimulus, which is why the query does a `LEFT JOIN` when fetching stimulus information.  Based on the data we have available, the assumption is that an Item will always have an `itemFilePath` but not necessarily a `stimulusFilePath`.

But the nullable `itembank.tblitem.filepath` trumps all that, so I'll add a `null` check.

millerts: This goes to the larger question if we should allow these to be nullable or not.  I think there are many fields in the client objects that are nullable.

jeffjohnson9046: I agree.  For this particular case, here's how I understand the item/stimulus files:
* You can have an item that does not have associated stimuli
* You cannot have an item without an item XML file

I've added the null check to the `itemFilePath`, because that's what the database says.  As I'm looking through the legacy code for implementing get page content, I'll check to see if having a null `itemFilePath` is an exceptional case.  If it is, we can come back and refactor the getter/setter to act accordingly.  How's that sound?

millerts: sgtm

millerts (approved):

client/src/main/java/tds/assessment/Item.java
jtreuting: I would argue that we could make the `tblitem.filepath` non-nullable or at least treat it as such in the application, since I can't think of a possible way that a valid item does not have an item path since without one you can't display an item in the TDS.

jtreuting (approved):

PR #25, opened 2017-01-10
Make utf-8 the default for the schemas

millerts:
There are four main changes:
* A V1 script that will set the charset and collation for the `configs` and `itembank` schemas since they are both used in tests.
* Added standard comments were missing in the migration files
* Added the constraints that were commented out back into the scripts
* Remove the engine and charset per table settings


service/src/test/resources/db/migration/beforeMigrate.sql
```diff
@@ -1,9 +0,0 @@
-/***********************************************************************************************************************
```
millerts, line 1: This isn't ever used by flyway.

jeffjohnson9046: I was just about to suggest that whatever's put in the V1 script to alter the databases could've also been put here, but there's no real difference.

millerts: I'll revisit this to see if we can get this working

jeffjohnson9046 (approved):

service/src/test/resources/db/migration/beforeMigrate.sql
millerts: @jeffjohnson9046 is right and I am wrong.  This file can be used rather than adding a V1 migration script.  I will modify this PR to include that change.

jtreuting (approved):

PR #26, opened 2017-01-17
Assessment Service - Caching + integration test coverage

burzum619:


service/src/main/java/tds/assessment/services/impl/AssessmentServiceImpl.java
```diff
@@ -40,8 +42,9 @@ public AssessmentServiceImpl(AssessmentQueryRepository assessmentQueryRepository
     }
 
     @Override
+    @Cacheable(CacheType.LONG_TERM)
     public Optional<Assessment> findAssessment(final String clientName, final String assessmentKey) {
-        Optional<Assessment> maybeAssessment = assessmentQueryRepository.findAssessmentByKey(assessmentKey, clientName);
+        Optional<Assessment> maybeAssessment = assessmentQueryRepository.findAssessmentByKey(clientName, assessmentKey);
```
burzum619, line 47: Noticed this bug while I was writing my integration test - parameters were switched up here.

service/src/test/java/tds/assessment/services/impl/AssessmentServiceImplIntegrationTest.java
```diff
@@ -0,0 +1,48 @@
...
+import tds.assessment.repositories.AssessmentQueryRepository;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+@RunWith(SpringJUnit4ClassRunner.class)
+@SpringBootTest
+public class AssessmentServiceImplIntegrationTest {
```
millerts: Looking at the pom.xml integration tests must be plural in the name.  So this should be renamed `AssessmentServiceImplIntegrationTests`.  You will notice this doesn't execute via Maven without the change.

millerts (changes requested):
One small change.

service/src/test/java/tds/assessment/services/impl/AssessmentServiceImplIntegrationTest.java
burzum619: Got it, I'll change and keep that in mind.

PR #27, opened 2017-01-17
Assessment Service - Field Test Item Group Selection 

burzum619:
This PR contains some Assessment service changes required for field test item group selection in TDS_ExamService. It includes the addition of some additional properties to the Segment and Item class, the implementation of equals() and hashCode() in Item, and some fieldtest-specific validation in the AssessmentAssembler.

### DIFF
--- a/client/src/main/java/tds/assessment/Item.java
+++ b/client/src/main/java/tds/assessment/Item.java
@@ -98,7 +98,8 @@ public class Item {
     }
 
     /**
-     * @return the block id of the item group
+     * @return the block id of the item group, used to group items into blocks for the field test selection algorithm.
+     * Typically "A" for most items.
      */
     public String getBlockId() {
         return blockId;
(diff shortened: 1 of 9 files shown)


client/src/main/java/tds/assessment/Item.java
```diff
@@ -97,6 +98,17 @@ public void setGroupKey(String groupKey) {
     }
 
     /**
+     * @return the block id of the item group
```
millerts: What is a block id?  And can we add what it is to this comment?

client/src/main/java/tds/assessment/Item.java
```diff
@@ -198,4 +210,33 @@ public void setPrintable(boolean printable) {
...
+        if (!groupKey.equals(item.groupKey)) return false;
+        if (!blockId.equals(item.blockId)) return false;
+        if (!strand.equals(item.strand)) return false;
+        if (itemProperties != null ? !itemProperties.equals(item.itemProperties) : item.itemProperties != null)
+            return false;
+        return formKeys != null ? formKeys.equals(item.formKeys) : item.formKeys == null;
+    }
+
+    @Override
+    public int hashCode() {
```
millerts, line 236: Just wanted to make sure you intended to make it possible for two items to not be equal but produce the same hashcode.

millerts: Note the last bullet point in the [hashcode](http://docs.oracle.com/javase/7/docs/api/java/lang/Object.html#hashCode()) contract.

client/src/main/java/tds/assessment/Segment.java
```diff
@@ -143,7 +168,7 @@ public void setFieldTestMinItems(int fieldTestMinItems) {
     /**
      * @return the maximum number of field test items in the {@link tds.assessment.Segment}
      */
-    public int getFieldTestMaxItems() {
+    public Integer getFieldTestMaxItems() {
```
millerts: Do these have to be nullable?  For example, is there harm if this is zero if null in the db?

millerts: Actually, as currently coded I don't think it is possible for them to be null because your using a primitive int to populate it ([here](https://github.com/SmarterApp/TDS_AssessmentService/pull/27/files#diff-855a89d888831240829f4e52cb2578cbR135)) and the autoboxing will not allow it to be null.

Same for all the field test ints.

millerts (changes requested):
Minor changes and clarifications requestd

client/src/main/java/tds/assessment/Segment.java
burzum619: In fact, it looks like the loader defaults these values to "0", so although its not enforced by the DB, I don't think it's totally fine to treats these as primitives. 

ftstartpos/ftendpos on the other hand are typically null for all assessments first loaded (even the ones that have min/max field test item counts defined).

burzum619: In the case of fieldTestStartPosition/EndPosition, I suppose although they are typically null in the `tblsetofadminsubjects` table, effectively it doesn't make much of a difference if they are null or zeros - in the case that there are ft min/maxes defined, we expect the start and end position to be non-null and non-zero, both would be Illegal States. And if there aren't any ft min/maxes defined, then we really don't care whats in the other field test columns. I think it's fine to make these primitives as well.

millerts: Cool.  The point I was trying to make with the link I provided was that if I followed the code correctly the mapper was autoboxing them to primitives anyway so I don't think the code as written would ever allow null regardless of hte object type here.

client/src/main/java/tds/assessment/Item.java
jtreuting: This is the note form the Admin Test Package specification:

"Used to group items into blocks for the field test selection algorithm.
Note that the Test Authoring assigns "A" for all items."

millerts: Oh yeah.  I remember that from the in person review.

service/src/test/java/tds/assessment/repositories/impl/AssessmentMapperTest.java
```diff
@@ -247,6 +261,44 @@ public void shouldReturnASegmentedAdaptiveAssessment() {
...
+                fieldTestStartDate,
+                fieldTestEndDate,
+                segmentFieldTestStartDate,
+                segmentFieldTestEndDate,
+                0,
+                0);
+
+        Optional<Assessment> maybeAssessment = mapper.mapResults(records);
+
+        assertThat(maybeAssessment).isPresent();
```
jtreuting: These `assertThat()` calls aren't needed right as it will throw the exception before it gets here.  Same for the previous test.

jtreuting (changes requested):
One more minor thing in the new tests

PR #28, opened 2017-01-22
Feature/accomodations in assessment

millerts:
There were some issues when communicating with assessment service from config service in docker.

The endpoints that were required assessment service always belonged in assessment service after we decided to build endpoints based on intent rather than strictly by DB.  Rather than investigate the docker issues I moved the endpoints since that was on the TODO list.

I need this for other work so I'll be merging this code but will make follow up changes depending on review.


service/src/main/java/tds/assessment/repositories/AccommodationsQueryRepository.java
```diff
@@ -0,0 +1,37 @@
...
+    List<Accommodation> findAccommodationsForNonSegmentedAssessmentByKey(String assessmentKey, Set<String> languageCodes);
+
+    /**
+     * Finds the accommodations for an assessment id an client
+     *
+     * @param clientName   the client name for the accommodations
+     * @param assessmentId the assessment id for the accommodations
+     * @return list of {@link tds.accommodation.Accommodation}
+     */
+    List<Accommodation> findAssessmentAccommodationsById(String clientName, String assessmentId);
```
jtreuting, line 36: Noticing that we aren't consistent with using `final` for parameters that don't change.  We should probably discuss in a parking lot at some point.

millerts: Agreed.  I'm fine using final.

PR #29, opened 2017-02-10
Changes to support checkApproval. 

burzum619:
Main things to note:

- Assessment now has a collection of tool dependencies. 
- Added the "Label" property to assessments and segments - these are used by the UI

### DIFF
--- a/client/src/main/java/tds/accommodation/Dependency.java
+++ b/client/src/main/java/tds/accommodation/Dependency.java
@@ -1,0 +1,141 @@
+package tds.accommodation;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+/**
+ * Represents an accommodation/tool dependency
+ */
+public class Dependency {
+    private String context;
+    private String ifType;
+    private String ifValue;
+    private String thenType;
+    private String thenValue;
+    
+    @JsonProperty("default")
+    private boolean isDefault;
+    
+    /**
+     * Empty constructor for frameworks
+     */
+    private Dependency() {}
+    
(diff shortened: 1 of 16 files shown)


client/src/main/java/tds/accommodation/Dependency.java
```diff
@@ -0,0 +1,141 @@
+package tds.accommodation;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+/**
+ * Represents an accommodation/tool dependency
+ */
+public class Dependency {
```
jtreuting: Dependency feels too generic.  Maybe `AccommodationDependency` or `ToolDependency`?

service/src/main/java/tds/assessment/repositories/AccommodationsQueryRepository.java
```diff
@@ -26,4 +27,13 @@
      * @return list of {@link tds.accommodation.Accommodation}
      */
     List<Accommodation> findAssessmentAccommodationsById(String clientName, String assessmentId);
+    
+    /**
+     * Finds the accommodation dependencies for an assessment id and clienet
```
jtreuting: typo `clienet` randomly caught my eye

jtreuting (approved):
Approved, though I have a recommendation for class name change to checkout

client/src/main/java/tds/accommodation/Dependency.java
burzum619: I vote AccommodationDependency in that case (since we already seem to be lumping "tools" into that category)

PR #43, opened 2017-03-20
Feature/fetch items

millerts:


service/src/test/java/tds/assessment/web/endpoints/ItemControllerIntegrationTests.java
```diff
@@ -0,0 +1,99 @@
...
+    private MockMvc http;
+
+    @Test
+    public void shouldReturnStimulusItemMetadata() throws Exception {
+        String url = "/assessments/item/metadata?clientName=SBAC_PT&bankKey=1&stimulusKey=3";
+
+        ItemFileMetadata itemFileMetadata = ItemFileMetadata.create(ItemFileType.STIMULUS, "1-3", "stimulusFile", "stimulusFile/");
+        when(mockItemService.findItemFileMetadataByStimulusKey("SBAC_PT", 1, 3)).thenReturn(Optional.of(itemFileMetadata));
+
+        MvcResult result = http.perform(get(url)
```
millerts, line 49: This is a different way of doing our controller integration tests.  As @burzum619 commented last week it is a bit of a pain that we don't verify that the client objects can be parsed by Jackson.  So I figured I'd try to see if we could tweak our integration test style to accommodate that complaint, and I think this works pretty well.

1. Inject the ObjectMapper that is created `JacksonObjectMapperConfiguration`
2. Use the `MockMvc` like always except instead of running assertions against the JSON body content return the `MvcResult`.
3. Use the objectMapper to read the content value (from `MvcResult.getResponse().getContentAsByteArray()`.
4. Do assertions on the translated object.

The `ObjectMapper` injected is the same one used by the RestTemplate in our other services.  This should reduce the number of times we run into the json parsing errors though of course if the caller doesn't configure the object mapper appropriately it'll still blow up but I see no way around that.

service/src/main/java/tds/assessment/repositories/impl/ItemQueryRepositoryImpl.java
```diff
@@ -138,4 +142,71 @@ public ItemQueryRepositoryImpl(final NamedParameterJdbcTemplate jdbcTemplate) {
         );
     }
 
+    @Override
+    public Optional<ItemFileMetadata> findItemFileMetadataByStimulusKey(final String clientName, final long bankKey, final long stimulusKey) {
```
millerts, line 146: I went back and forth on modifying the existing query, but I hesitate to do that until all of the scoring stuff has shaken out.

service/src/test/java/tds/assessment/web/endpoints/ItemControllerIntegrationTests.java
burzum619: Nice, I like this approach.

burzum619 (approved):

client/src/main/java/tds/assessment/ItemFileMetadata.java
```diff
@@ -0,0 +1,37 @@
...
+     * @return the file path
+     */
+    public abstract String getFilePath();
+
+    @JsonCreator
+    public static ItemFileMetadata create(
+        @JsonProperty("itemType") final ItemFileType newItemType,
+        @JsonProperty("id") final String newId,
+        @JsonProperty("fileName") final String newFileName,
+        @JsonProperty("filePath") final String newFilePath) {
```
wtritch, line 34: It still weirds me out that we're using filePaths rather than Blobs to hold the data...

millerts: We are hamstrung by the existing system.  We're walking a balance between improving what we can in the time provided without rewriting all of it.

service/src/main/java/tds/assessment/web/endpoints/ItemController.java
```diff
@@ -0,0 +1,50 @@
...
+        this.itemService = itemService;
+    }
+
+    @GetMapping(value = "assessments/item/metadata", produces = MediaType.APPLICATION_JSON_VALUE)
+    @ResponseBody
+    ResponseEntity<ItemFileMetadata> findItemFileMetadata(@RequestParam final String clientName,
+                                                          @RequestParam final long bankKey,
+                                                          @RequestParam(required = false) final Long stimulusKey,
+                                                          @RequestParam(required = false) final Long itemKey) {
+        if (stimulusKey == null && itemKey == null) {
```
wtritch, line 32: Maybe isBlank checks?

millerts: We only care if both are not provided since one or the other is required.  If there is a better way to construct the URL we can tweak it.

And if they aren't provided by the caller they'll be null.  Blank is fine it'll just result in a 404.

service/src/test/java/tds/assessment/web/endpoints/ItemControllerIntegrationTests.java
wtritch: I like it.  But to test that our API returns as expected, I still think testing using jsonPath is useful.  (e.g. we could configure our ObjectMapper to output any property as any name and round-tripping would work, but the client would break.)

Not requesting a change, but I wanted to throw out a sample controller test from my last project and see what you guys thought:
```
@Test
    public void itShouldCreateABatchComment() throws Exception {
        final Batch batch = createBatch(mock(Vendor.class), UUID.randomUUID().toString());
        when(this.batchService.findOne(eq(batch.getId()))).thenReturn(batch);

        final Comment postedComment = new CommentBuilder()
                .withMessage("This is a comment")
                .build();

        final User authUser = createUser("userId", "A User");
        final Comment persistedComment = createComment(batch, authUser, UUID.randomUUID().toString());
        when(this.commentService.post(any(Comment.class), eq(batch), any(User.class))).thenReturn(persistedComment);
        asUser(authUser);
        try {
            getHttp().perform(post(fromRoot("/batches/" + batch.getId() + "/comments"))
                    .content(getObjectMapper().writeValueAsString(postedComment))
                    .contentType(APPLICATION_JSON)
                    .accept(APPLICATION_JSON))

                    .andExpect(status().is(CREATED.value()))
                    .andExpect(jsonPath("$.id", equalTo(persistedComment.getId())))
                    .andExpect(jsonPath("$.createdDate", instantAs(persistedComment.getCreatedDate())))
                    .andExpect(jsonPath("$.message", equalTo(persistedComment.getMessage())))
                    .andExpect(jsonPath("$.user.name", equalTo(persistedComment.getUser().getName())))
                    .andExpect(jsonPath("$.user.iconUrl", equalTo(persistedComment.getUser().getIconUrl())))
                    .andExpect(jsonPath("$.user.email", equalTo(persistedComment.getUser().getEmail())));

            final ArgumentCaptor<Comment> commentCaptor = ArgumentCaptor.forClass(Comment.class);
            final ArgumentCaptor<User> userCaptor = ArgumentCaptor.forClass(User.class);
            verify(this.commentService).post(commentCaptor.capture(), eq(batch), userCaptor.capture());
            assertThat(commentCaptor.getValue().getMessage()).isEqualTo(postedComment.getMessage());
            assertThat(userCaptor.getValue().getId()).isEqualTo(authUser.getId());
        } finally {
            clearUser();
        }
    }
```

wtritch (approved):

service/src/main/java/tds/assessment/web/endpoints/ItemController.java
wtritch: awesome

client/src/main/java/tds/assessment/ItemFileMetadata.java
wtritch: I'm just not looking forward to setting up an S3 shared volume in our container images =P

PR #44, opened 2017-03-21
Missing IntraTest TTS Controls accommodation

burzum619:
Adding ‘online’ testmode to assessment accommodations query. This is the test mode for ‘IntraTest TTS Controls’, the only accommodation of this "testmode" value.

https://jira.fairwaytech.com/browse/TDS-790

### DIFF
--- a/service/src/main/java/tds/assessment/repositories/impl/AccommodationsQueryRepositoryImpl.java
+++ b/service/src/main/java/tds/assessment/repositories/impl/AccommodationsQueryRepositoryImpl.java
@@ -282,7 +282,7 @@ public class AccommodationsQueryRepositoryImpl implements AccommodationsQueryRep
                 "  and TType.Context = '*' \n" +
                 "  and TT.ContextType = 'TEST' \n" +
                 "  and TT.Context = '*' \n" +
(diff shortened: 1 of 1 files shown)


service/src/main/java/tds/assessment/repositories/impl/AccommodationsQueryRepositoryImpl.java
```diff
@@ -282,7 +282,7 @@ public AccommodationsQueryRepositoryImpl(final NamedParameterJdbcTemplate jdbcTe
                 "  and TType.Context = '*' \n" +
                 "  and TT.ContextType = 'TEST' \n" +
                 "  and TT.Context = '*' \n" +
-                "  and (TType.TestMode = 'ALL' AND TT.TestMode = 'ALL') \n" +
+                "  and (TType.TestMode IN ('ALL', 'online') AND TT.TestMode = 'ALL') \n" +
```
jtreuting, line 285: It doesn't need to be on both of the TestMode checks?

burzum619: Nope, for some reason in the seed data that inserts the TTS IntraTest Control record, the testmode is set to 'online' only in testtooltype. testtool is still 'ALL'

PR #45, opened 2017-03-25
Optimization - Assessment Accommodations -Moving client name filter into WHERE and out of JOIN ON. 

burzum619:
Reduced total query execution times from 19 seconds - ~800 ms. This call has a long-term cache.


jtreuting (approved):
Nice!

PR #52, opened 2017-04-12
Adding grades and other properties to Assessment used by TRT

burzum619:


millerts (approved):
I understand this is heavily cached, but the data we just added are only for reporting purposes.  Not sure if it needs to be returned on every assessment query.  Not asking for any changes.  Just something to think about as we continue to grow this object, which is already the largest object in the stack.

millerts (commented):
I don't see any changes to the Controller integration tests to account for these new fields.  Please add those as well.

Or we should ditch them.

burzum619:
Agreed. Added controller integration test coverage with a fully randomized Assessment object. Also added equals/hashcode to many of the models in Assessment that didn't have them.

PR #54, opened 2017-04-12
Feature/item selection

millerts:
This represents the code needed to port the existing item selection data access code found [here](https://github.com/SmarterApp/TDS_TestDeliverySystemDataAccess/blob/master/tds-dll-mysql/src/main/java/tds/dll/mysql/ItemSelectionDLL.java#L2377).


client/src/main/java/tds/assessment/Item.java
```diff
@@ -24,7 +24,7 @@
     private boolean fieldTest;
     private boolean required;
     private String strand;
-    private List<ItemProperty> itemProperties;
+    private List<ItemProperty> itemProperties = new ArrayList<>();
```
burzum619, line 27: No issue with this, but I know that in other model classes such as Assessment and Segment, we initialize the empty array in the getter. Don't feel very strongly about it in either direction, but if we care about consistency this can follow that pattern.

service/src/main/java/tds/assessment/repositories/AssessmentQueryRepository.java
```diff
@@ -29,4 +29,5 @@
      */
     List<ItemConstraint> findItemConstraintsForAssessment(final String clientName, final String assessmentId);
 
+    Optional<Assessment> findAssessmentBySegmentKey(final String segmentKey);
```
burzum619, line 32: javadocs?

service/src/main/java/tds/assessment/repositories/StrandQueryRepository.java
```diff
@@ -16,4 +18,6 @@
      * @return the list of the strands for the assessment and its segments
      */
     Set<Strand> findStrands(final String assessmentKey);
+
+    List<ContentLevelSpecification> findContentLevelSpeficationsBySegmentKey(final String segmentKey);
```
burzum619, line 22: javadocs

client/src/main/java/tds/assessment/Item.java
wtritch: If we want true immutability this could be `final`, the getter could return an ImmutableList.copyOf(...) and the setter could clear and add all...

service/src/main/java/tds/assessment/repositories/impl/AssessmentMapper.java
```diff
@@ -156,6 +156,30 @@ private Segment getSegment(final ResultRow row) {
         segment.setFieldTestMaxItems(ftMaxItems);
         segment.setFieldTestStartPosition(ftStartPosition);
         segment.setFieldTestEndPosition(ftEndPosition);
+
```
wtritch, line 159: nit: extra space

service/src/main/java/tds/assessment/repositories/impl/ItemMeasurementQueryRepositoryImpl.java
```diff
@@ -0,0 +1,65 @@
...
+        this.jdbcTemplate = jdbcTemplate;
+    }
+
+    @Override
+    public List<ItemMeasurement> findItemMeasurements(final String segmentKey, final String assessmentKey) {
+        SqlParameterSource parameters = new MapSqlParameterSource("segmentKey", segmentKey)
+            .addValue("assessmentKey", assessmentKey);
+
+        String SQL = "SELECT \n" +
+            "   coalesce(S.TestPosition, 1) as segmentPosition, \n" +
```
wtritch, line 29: ew.  We have to define defaults at selection time?

service/src/main/java/tds/assessment/repositories/impl/StrandQueryRepositoryImpl.java
```diff
@@ -43,15 +55,88 @@ public StrandQueryRepositoryImpl(final NamedParameterJdbcTemplate jdbcTemplate)
...
+            "   S.abilityWeight, \n" +
+            "   S.precisionTarget, \n" +
+            "   S.precisionTargetMetWeight, \n" +
+            "   S.precisionTargetNotMetWeight, \n" +
+            "   CASE WHEN SS._fk_Parent IS NULL THEN 0 ELSE 1 END AS elementType \n" +
+            "FROM tbladminstrand S  \n " +
+            "JOIN tblstrand SS ON SS._Key = S._fk_Strand \n" +
+            "WHERE S._fk_AdminSubject = :segmentKey ) \n" +
+            "UNION ALL \n" +
+            "(SELECT \n" +
```
wtritch, line 86: @jeffjohnson9046 has run into issues with MySql 5.6 and sub-selects before, although I don't remember the details.  Just wanted to make sure this is gtg?

burzum619 (approved):

wtritch (approved):

service/src/main/java/tds/assessment/repositories/impl/StrandQueryRepositoryImpl.java
millerts: We've been doing sub selects prior to any of the issues @jeffjohnson9046 hit.  However, I ran these queries against dev in my testing so I think we're good.  In addition, these queries are almost all straight ports of existing legacy queries.

service/src/main/java/tds/assessment/repositories/impl/ItemMeasurementQueryRepositoryImpl.java
millerts: I am porting legacy queries.  At this point in time I'd rather not shift around logic and just leave the queries as is unless we find performance issues.

Or we get done early and are willing to take the time to revisit the interesting logic in this area.

PR #61, opened 2017-04-24
Adding AssessmentInfo controller - ability to obtain information on m…

burzum619:
…ultiple assessments in one repo call

This API will be called on by ExamService in order  to get Assessment metadata for multiple assessments, required for checking exam eligibility.

### DIFF
--- a/client/src/main/java/tds/assessment/AssessmentInfo.java
+++ b/client/src/main/java/tds/assessment/AssessmentInfo.java
@@ -1,0 +1,186 @@
+package tds.assessment;
+
+import java.util.List;
+
+/**
+ * An object containing assessment metadata
+ */
+public class AssessmentInfo {
+    private String key;
+    private String id;
+    private String subject;
+    private String label;
+    private int maxAttempts;
+    private List<String> grades;
+    private List<String> languages;
+    private List<AssessmentWindow> assessmentWindows;
+
+    /**
+     * Empty constructor for frameworks
+     */
+    private AssessmentInfo() {}
+
+    public AssessmentInfo(Builder builder) {
+        this.maxAttempts = builder.maxAttempts;
+        this.assessmentWindows = builder.assessmentWindows;
+        this.id = builder.id;
+        this.label = builder.label;
+        this.subject = builder.subject;
+        this.grades = builder.grades;
+        this.languages = builder.languages;
+        this.key = builder.key;
+    }
+
+    public static final class Builder {
+        private String key;
+        private String id;
+        private String subject;
+        private String label;
+        private int maxAttempts;
+        private List<String> grades;
+        private List<String> languages;
+        private List<AssessmentWindow> assessmentWindows;
+
+        public Builder withKey(String key) {
+            this.key = key;
+            return this;
+        }
+
+        public Builder withId(String id) {
+            this.id = id;
+            return this;
+        }
+
+        public Builder withSubject(String subject) {
+            this.subject = subject;
+            return this;
+        }
+
+        public Builder withLabel(String label) {
+            this.label = label;
+            return this;
+        }
+
+        public Builder withMaxAttempts(int maxAttempts) {
+            this.maxAttempts = maxAttempts;
+            return this;
+        }
+
+        public Builder withGrades(List<String> grades) {
+            this.grades = grades;
+            return this;
+        }
+
+        public Builder withLanguages(List<String> languages) {
+            this.languages = languages;
+            return this;
+        }
+
+        public Builder withAssessmentWindows(List<AssessmentWindow> assessmentWindows) {
+            this.assessmentWindows = assessmentWindows;
+            return this;
+        }
+
+        public Builder fromAssessmentInfo(AssessmentInfo assessmentInfo) {
+            this.maxAttempts = assessmentInfo.maxAttempts;
+            this.assessmentWindows = assessmentInfo.assessmentWindows;
+            this.id = assessmentInfo.id;
+            this.label = assessmentInfo.label;
+            this.subject = assessmentInfo.subject;
+            this.grades = assessmentInfo.grades;
+            this.languages = assessmentInfo.languages;
+            this.key = assessmentInfo.key;
+            return this;
+        }
+
+        public AssessmentInfo build() {
+            return new AssessmentInfo(this);
+        }
+    }
+
+    /**
+     * @return The key of the assessment
+     */
+    public String getKey() {
+        return key;
+    }
+
+    /**
+     * @return The id of the assessment
+     */
+    public String getId() {
+        return id;
+    }
+
+    /**
+     * @return The subject of the assessment
+     */
+    public String getSubject() {
+        return subject;
+    }
+
+    /**
+     * @return The label of the assessment, displayed in the UI
+     */
+    public String getLabel() {
+        return label;
+    }
+
+    /**
+     * @return The maximum number of exam attempts
+     */
+    public int getMaxAttempts() {
+        return maxAttempts;
+    }
+
+    /**
+     * @return The list of eligible grades for the assessments
+     */
+    public List<String> getGrades() {
+        return grades;
+    }
+
+    /**
+     * @return The list of languages for the assessment
+     */
+    public List<String> getLanguages() {
+        return languages;
+    }
+
+    /**
+     * @return The assessment windows for the assessment
+     */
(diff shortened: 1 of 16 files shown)


service/src/main/java/tds/assessment/repositories/impl/AssessmentQueryRepositoryImpl.java
```diff
@@ -147,6 +149,51 @@ public AssessmentQueryRepositoryImpl(final NamedParameterJdbcTemplate jdbcTempla
...
+
+        final String SQL =
+            "SELECT \n" +
+                "   a._key AS assessmentKey, \n" +
+                "   a.testid AS assessmentId, \n" +
+                "   tp.subjectname AS subject, \n" +
+                "   tp.label AS assessmentLabel, \n" +
+                "   GROUP_CONCAT(DISTINCT tool.code) languages, \n" +
+                "   GROUP_CONCAT(DISTINCT g.grade) grades, \n" +
+                "   tp.maxopportunities AS maxAttempts \n" +
```
burzum619, line 166: The GROUP_CONCAT function is not something I had used before - it groups multiple rows of values into one comma delimited list (for example, an assessment with grades "6" and "7" would end up with the grades column as "6,7". We then parse this in the RowMapper with a simple .split(","). I thought this might be a bit of a simpler approach then some of the other solutions we are currently using, but feel free to point out any flaws with this.

client/src/main/java/tds/assessment/AssessmentInfo.java
```diff
@@ -0,0 +1,186 @@
...
+ */
+public class AssessmentInfo {
+    private String key;
+    private String id;
+    private String subject;
+    private String label;
+    private int maxAttempts;
+    private List<String> grades;
+    private List<String> languages;
+    private List<AssessmentWindow> assessmentWindows;
```
wtritch: Should these be initialized to non-null collections here?

client/src/main/java/tds/assessment/AssessmentInfo.java
```diff
@@ -0,0 +1,186 @@
...
+        public Builder fromAssessmentInfo(AssessmentInfo assessmentInfo) {
+            this.maxAttempts = assessmentInfo.maxAttempts;
+            this.assessmentWindows = assessmentInfo.assessmentWindows;
+            this.id = assessmentInfo.id;
+            this.label = assessmentInfo.label;
+            this.subject = assessmentInfo.subject;
+            this.grades = assessmentInfo.grades;
+            this.languages = assessmentInfo.languages;
+            this.key = assessmentInfo.key;
+            return this;
```
wtritch: Do we want any validation in here? (null checks, etc)

client/src/main/java/tds/assessment/AssessmentInfo.java
```diff
@@ -0,0 +1,186 @@
...
+            this.languages = languages;
+            return this;
+        }
+
+        public Builder withAssessmentWindows(List<AssessmentWindow> assessmentWindows) {
+            this.assessmentWindows = assessmentWindows;
+            return this;
+        }
+
+        public Builder fromAssessmentInfo(AssessmentInfo assessmentInfo) {
```
millerts: Can we make this static and return a new builder?  `public static Builder fromAssessmentInfo(info)`

service/src/main/java/tds/assessment/repositories/impl/AssessmentQueryRepositoryImpl.java
```diff
@@ -147,6 +149,51 @@ public AssessmentQueryRepositoryImpl(final NamedParameterJdbcTemplate jdbcTempla
...
+
+        return jdbcTemplate.query(SQL, parameters, (rs, row) ->
+            new AssessmentInfo.Builder()
+                .withKey(rs.getString("assessmentKey"))
+                .withId(rs.getString("assessmentId"))
+                .withLabel(rs.getString("assessmentLabel"))
+                .withSubject(rs.getString("subject"))
+                .withMaxAttempts(rs.getInt("maxAttempts"))
+                .withLanguages(Arrays.asList(rs.getString("languages").split(",")))
+                .withGrades(Arrays.asList(rs.getString("grades").split(",")))
```
wtritch: In general I prefer using a Guava Splitter:
```
private static final Splitter COMMA_SPLITTER = Splitter.on(",");

.withGrades(COMMA_SPLITTER.splitToList(rs.getString("grades"));
```
Because it simplifies a lot of splitting logic.  But personal preference =)

service/src/main/java/tds/assessment/services/impl/AssessmentInfoServiceImpl.java
```diff
@@ -0,0 +1,44 @@
...
+    public AssessmentInfoServiceImpl(final AssessmentQueryRepository assessmentQueryRepository,
+                                     final AssessmentWindowService assessmentWindowService) {
+        this.assessmentQueryRepository = assessmentQueryRepository;
+        this.assessmentWindowService = assessmentWindowService;
+    }
+
+    @Override
+    public List<AssessmentInfo> findAssessmentInfo(final String clientName, final String... assessmentKeys) {
+        List<AssessmentInfo> assessments = assessmentQueryRepository.findAssessmentInfoByKeys(clientName, assessmentKeys);
+        Set<String> assessmentIds = assessments.stream().map(AssessmentInfo::getId).collect(Collectors.toSet());
```
millerts: I think you can use `toArray` on the stream.  And if uniqueness is important you can do a `distinct()` after the map.

service/src/main/java/tds/assessment/repositories/impl/AssessmentQueryRepositoryImpl.java
wtritch: None of the values will ever, ever have a comma in them?

millerts (approved):
Couple small suggestions

service/src/main/java/tds/assessment/services/impl/AssessmentInfoServiceImpl.java
```diff
@@ -0,0 +1,44 @@
...
+        Map<String, List<AssessmentWindow>> assessmentWindows = assessmentWindowService.findAssessmentWindowsForAssessmentIds(
+            clientName, assessmentIds.toArray(new String[assessmentIds.size()]));
```
wtritch: I have to admit, this feels a little funky.  If it's going to be called via code, I'd recommend switching the findAssessmentWindowsForAssessmentIds method to accept a collection rather than a varargs param.

service/src/main/java/tds/assessment/services/AssessmentInfoService.java
```diff
@@ -0,0 +1,19 @@
...
+ */
+public interface AssessmentInfoService {
+    /**
+     * Fetches the list of {@link tds.assessment.AssessmentInfo} for a given collection of assessmentKeys
+     *
+     * @param clientName     The client name of the TDS environment
+     * @param assessmentKeys The collection of keys to obtain {@link tds.assessment.AssessmentInfo} for
+     * @return A list of {@link tds.assessment.AssessmentInfo}
+     */
+    List<AssessmentInfo> findAssessmentInfo(final String clientName, final String... assessmentKeys);
```
wtritch, line 18: Nit: if order is not important or guaranteed, I generally prefer returning/accepting Collection<Blah> rather than List<Blah> in my interfaces.

service/src/main/java/tds/assessment/repositories/impl/AssessmentQueryRepositoryImpl.java
burzum619: Good point. I don't believe so, not in practice anyway. Theoretically its possible to insert any value into that VARCHAR column ("kin,dergar,den"), but I do not thing this is something to  be seriously concerned about.

wtritch (approved):
LGTM

service/src/main/java/tds/assessment/services/impl/AssessmentInfoServiceImpl.java
burzum619: Agreed, I'm cool with changing findAssessmentWindowsForAssessmentIds to accept a Set instead.

wtritch: \<cough\>Collection\<cough\> =)

PR #64, opened 2017-04-26
Adding API to fetch AssessmentInfos by grade

burzum619:
This API will be used by the exam service on guest logins

### DIFF
--- a/service/src/main/java/tds/assessment/repositories/impl/AssessmentQueryRepositoryImpl.java
+++ b/service/src/main/java/tds/assessment/repositories/impl/AssessmentQueryRepositoryImpl.java
@@ -4,11 +4,15 @@ import com.google.common.base.Splitter;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.jdbc.core.RowMapper;
 import org.springframework.jdbc.core.namedparam.MapSqlParameterSource;
 import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
 import org.springframework.jdbc.core.namedparam.SqlParameterSource;
 import org.springframework.stereotype.Repository;
 
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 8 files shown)


service/src/main/java/tds/assessment/repositories/impl/AssessmentQueryRepositoryImpl.java
```diff
@@ -182,17 +186,42 @@ public AssessmentQueryRepositoryImpl(final NamedParameterJdbcTemplate jdbcTempla
...
+        return jdbcTemplate.query(SQL, parameters, new AssessmentInfoRowMapper());
+    }
+
+    @Override
+    public List<AssessmentInfo> findAssessmentInfoForGrade(final String clientName, final String grade) {
+        final SqlParameterSource parameters = new MapSqlParameterSource("grade", grade)
+            .addValue("clientName", clientName);
+
+        final String SQL =
+            "SELECT \n" +
```
millerts: Suggest pull this select, from, and joins into a constant method that the two methods share.  That way if we need to change the mapper for whatever reason we can change the query in one place and the mapper.

At least it looks like the only difference is the where and group by.

service/src/main/java/tds/assessment/web/endpoints/AssessmentInfoController.java
```diff
@@ -26,9 +26,15 @@ public AssessmentInfoController(final AssessmentInfoService assessmentInfoServic
...
 
-    @GetMapping(value = "/{clientName}/assessments/", produces = MediaType.APPLICATION_JSON_VALUE)
+    @GetMapping(value = "/{clientName}/assessments", produces = MediaType.APPLICATION_JSON_VALUE)
     @ResponseBody
-    ResponseEntity<List<AssessmentInfo>> findAssessment(@PathVariable final String clientName, @RequestParam final String... assessmentKeys) {
+    ResponseEntity<List<AssessmentInfo>> findAssessmentInfos(@PathVariable final String clientName, @RequestParam final String... assessmentKeys) {
         return ResponseEntity.ok(assessmentInfoService.findAssessmentInfo(clientName, assessmentKeys));
     }
+
+    @GetMapping(value = "/{clientName}/assessments/grade/{grade}", produces = MediaType.APPLICATION_JSON_VALUE)
```
millerts: Why have a separate endpoint for this since it returns the same resource?  It seems like the existing endpoint should have an optional query parameter that will narrow it by grade.

`/SBAC_PT/assessments?grade=3`

millerts (approved):
Outside the endpiont comment LGTM

PR #68, opened 2017-04-28
Feature/add item groups to assessment

millerts:
This is needed to fix item selection.


client/src/main/java/tds/assessment/Segment.java
```diff
@@ -55,6 +55,7 @@
     private float slope;
     private float intercept;
     private float startInfo;
+    private List<ItemGroup> itemGroups;
```
wtritch: Initialize to empty list?  Or return an empty list in the getter if it's null?

client/src/main/java/tds/assessment/Segment.java
```diff
@@ -558,6 +562,18 @@ public void setIntercept(final float intercept) {
         this.intercept = intercept;
     }
 
+    /**
+     * @return The item groups that are associated with this segment.  This can be empty if there are not groups.  In
+     * those instances the item will be given a hard coded group in the selection code.
+     */
+    public List<ItemGroup> getItemGroups() {
```
millerts, line 569: This is the addition

wtritch (approved):

PR #70, opened 2017-04-29
Get all items regardless of form

millerts:
The crux of the issue is that items can be used across segment keys.   For example, @wtritch and I found an item that is used on a Math 7 and a Math 11 assessment.  This PR not only fixes the query, but it also simplifies the item fetching allowing the assembler to have all the business logic


client/src/main/java/tds/assessment/ItemProperty.java
```diff
@@ -10,18 +10,10 @@
     private String description;
 
     public ItemProperty(String name, String value, String description, String itemId) {
```
millerts, line 12: These other ones didn't make sense because if item id isn't populated other places in the system throw NPE's.

service/src/main/java/tds/assessment/repositories/impl/ItemQueryRepositoryImpl.java
```diff
@@ -69,6 +66,7 @@ public ItemQueryRepositoryImpl(final NamedParameterJdbcTemplate jdbcTemplate) {
                 "JOIN \n" +
                 "   itembank.testformitem formItem \n" +
                 "   ON formItem._fk_item = item._key \n" +
+                "   AND formItem._fk_adminsubject = adminItems._fk_adminsubject \n" +
```
millerts, line 69: This is one of the bugs.  Items can be referenced on many forms across segments.  We weren't isolating that.

service/src/main/java/tds/assessment/repositories/impl/ItemQueryRepositoryImpl.java
```diff
@@ -97,48 +95,37 @@ public ItemQueryRepositoryImpl(final NamedParameterJdbcTemplate jdbcTemplate) {
...
-                item.setbVector(resultExtractor.getString("bVector"));
-                item.setItemResponseTheoryModel(resultExtractor.getString("irtModel"));
-                item.setItemResponseTheoryAParameter((Float) resultExtractor.getObject("irtA"));
-                item.setItemResponseTheoryBParameter(resultExtractor.getString("irtB"));
-                item.setItemResponseTheoryCParameter((Float) resultExtractor.getObject("irtC"));
-                item.setClaims(resultExtractor.getString("clString"));
-                item.setItemKey(resultExtractor.getLong("itemKey"));
-                item.setBankKey(resultExtractor.getLong("bankKey"));
-            }
+        return jdbcTemplate.query(itemsSQL, parameters, (rs, rowNum) -> {
```
millerts, line 98: So this really shouldn't happen much anymore because we're not getting item records across segments which is what was happening prior to the query change above.  That being said it seemed like his had a bit too much logic when trying to collapse the items.  Instead, we'll return all the items and allow the assembler to handle it.

service/src/main/java/tds/assessment/services/impl/AssessmentAssembler.java
```diff
@@ -53,29 +53,29 @@ static void assemble(final Assessment assessment,
...
-        
+
         // FIXED FORM SEGMENTS - If forms were supplied, wire up the items to their appropriate form
         // RULE:  The items in the form should be sorted by their formPosition in ascending order
         if (forms.size() > 0) {
             for (Form form : forms) {
                 List<Item> itemsForThisForm = items.stream()
-                    .filter(i -> i.getFormKeys() != null
-                        && i.getFormKeys().contains(form.getKey())
+                    .filter(i -> i.getFormKey() != null
```
millerts, line 71: Since we're not collapsing the items in the repository we simply iterate over all of them and assign them to the proper form.

service/src/main/java/tds/assessment/services/impl/AssessmentAssembler.java
```diff
@@ -84,12 +84,13 @@ static void assemble(final Assessment assessment,
...
         }
-        
+
         // ADAPTIVE SEGMENTS - Update the segment with data based on its selection algorithm
         for (Segment segment : assessment.getSegments()) {
             segment.setStrands(strands.stream()
                 .filter(strand -> strand.getSegmentKey().equals(segment.getKey()))
                 .collect(Collectors.toSet()));
+
             segment.setItems(items.stream()
```
millerts, line 94: Ok, I tried to remove setting the items for fixed form since all the items are on the form, but that change ripples.  I'll remove the extra item setting in a separate PR because it isn't critical to remove the blocker for scoring.

service/src/test/java/tds/assessment/services/impl/AssessmentAssemblerTest.java
```diff
@@ -19,1018 +17,99 @@
...
-import static io.github.benas.randombeans.api.EnhancedRandom.randomListOf;
 import static org.assertj.core.api.Assertions.assertThat;
 
 public class AssessmentAssemblerTest {
     @Test
     public void shouldAssembleASegmentedFixedFormAssessment() {
-        // Assessment Init
-        final String assessmentId = "foo";
         Assessment assessment = new Assessment();
-        assessment.setKey("theKey");
```
millerts, line 31: These tests were hard to follow, and I struggled with modifying them after the logic changes.  After some analysis it seems like a lot of the data created for these tests doesn't actually check a unique logic block. 

Instead of modifying all the data I created test data that would hit all the different logic blocks simplifying it for future maintainers.

jeffjohnson9046 (approved):

PR #75, opened 2017-05-11
TDS-880 - Item stimuli paths should be null, not empty string

burzum619:


client/src/main/java/tds/assessment/Item.java
```diff
@@ -195,7 +195,7 @@ public String getItemFilePath() {
 
     public void setItemFilePath(String itemFilePath) {
         if (itemFilePath == null) {
-            itemFilePath = "";
+            throw new IllegalArgumentException("Item file path cannot be null");
```
millerts: I think you can use itemFilePath = checkNotNull(itemFilepath);  In common legacy there is a Preconditions class.

millerts (approved):
small tweak otherwise LGTM

PR #78, opened 2017-05-19
Handle adaptive items

millerts:


service/src/main/java/tds/assessment/repositories/impl/ItemQueryRepositoryImpl.java
```diff
@@ -63,7 +63,7 @@ public ItemQueryRepositoryImpl(final NamedParameterJdbcTemplate jdbcTemplate) {
                 "JOIN \n" +
                 "   itembank.tblitem item \n" +
                 "   ON item._key = adminItems._fk_item \n" +
-                "JOIN \n" +
+                "LEFT JOIN \n" +
```
millerts, line 66: Adaptive items do not have forms and are not present in testformitem.  So whenever we tried to get items for an adaptive exam we were getting 0 items.

service/src/test/java/tds/assessment/repositories/impl/ItemQueryRepositoryImplIntegrationTests.java
```diff
@@ -82,10 +82,7 @@ public void setUp() {
                 "('(SBAC_PT)SBAC-SEG1-MATH-8-Spring-2013-2015', 528, 528, 'PracTest::MG8::S1::SP14', 'ENU', '187-528', NULL, 0, 8233, NULL, 'Default'),\n" +
                 "('(SBAC_PT)SBAC-SEG1-MATH-8-Spring-2013-2015', 529, 529, 'PracTest::MG8::S1::SP14::Braille', 'ENU-Braille', '187-529', NULL, 0, 8233, NULL, 'Default'),\n" +
                 "('(SBAC_PT)SBAC-SEG2-MATH-8-Spring-2013-2015', 532, 532, 'PracTest::MG8::S2::SP14::Braille', 'ENU', '187-532', NULL, 0, 8233, NULL, 'Default'),\n" +
-                "('(SBAC_PT)SBAC-SEG2-MATH-8-Spring-2013-2015', 534, 534, 'PracTest::MG8::S2::SP14', 'ENU-Braille', '187-534', NULL, 0, 8233, NULL, 'Default'),\n" +
```
millerts, line 85: The test data was set up wrong which is why we didn't see this in our testing.  Basically, we were giving adaptive tests forms and form items.  Removed those so the (SBAC_PT)IRP-Perf-ELA-11-Summer-2015-2016 is the adaptive looking assessment

service/src/test/java/tds/assessment/repositories/impl/ItemQueryRepositoryImplIntegrationTests.java
```diff
@@ -187,23 +181,13 @@ public void shouldRetrieveActiveItemPropsForNonSegAssesssment() {
...
-    public void shouldRetrieveSingleItemNonSegmentedTest() {
+    public void shouldRetrieveSingleItemNonSegmentedAdaptiveTest() {
         List<Item> items = repository.findItemsForAssessment("(SBAC_PT)IRP-Perf-ELA-11-Summer-2015-2016");
-        assertThat(items).hasSize(2);
-
-        Item item1 = null;
-        Item item2 = null;
+        assertThat(items).hasSize(1);
 
-        for (Item item : items) {
```
millerts, line 197: I removed the second item from the assessment test data and didn't feel like adding it back in since I didn't feel it added much value.

service/src/main/java/tds/assessment/repositories/impl/ItemQueryRepositoryImpl.java
```diff
@@ -93,7 +93,7 @@ public ItemQueryRepositoryImpl(final NamedParameterJdbcTemplate jdbcTemplate) {
                 "AND \n" +
                 "   adminItems.isactive = 1 \n" +
                 "AND \n" +
-                "   formItem.isactive = 1";
+                "   (formItem.isactive = 1 OR formItem.isactive IS NULL)";
```
jeffjohnson9046, line 96: The `LEFT JOIN` totally makes sense, but this is a bit confusing to me.  Is the idea here that if we _are_ looking for forms that we get the active and inactive (or whatever a `NULL isactive` field means) forms and let the caller decide what to do with the result?  I'm sure it'll work fine, just seems like an odd condition to me.

jeffjohnson9046 (approved):

service/src/main/java/tds/assessment/repositories/impl/ItemQueryRepositoryImpl.java
jtreuting: the `isactive IS NULL` is because of the `LEFT JOIN` would have `NULL`s for all the columns in testformitem

jtreuting (approved):

PR #79, opened 2017-06-01
Bugfix/fix sibling items query

millerts:


client/src/main/java/tds/assessment/SegmentItemInformation.java
```diff
@@ -47,10 +47,10 @@ public Segment getSegment() {
     }
 
     /**
-     * @return the parent assessment's {@link tds.assessment.Item}.  If a single segmented assessment this will be empty
+     * @return the assessment's {@link tds.assessment.Item}.  If a single segmented assessment this will be empty
      */
-    public List<Item> getParentItems() {
-        return parentItems;
+    public List<Item> getSiblingItems() {
```
millerts, line 52: This was a misunderstanding on my part.  This needs to contain any item that isn't part of the current segment.  I mistakenly thought it was the parent items which never really happens in a multi segmented assessment.

service/src/main/java/tds/assessment/repositories/impl/AssessmentQueryRepositoryImpl.java
```diff
@@ -245,52 +248,30 @@ public AssessmentQueryRepositoryImpl(final NamedParameterJdbcTemplate jdbcTempla
...
-                "   ON TA._fk_client = CL._key \n" +
-                "WHERE \n" +
-                "   B._key = :segmentKey \n OR A._key = :segmentKey \n" +
-                "ORDER BY  \n" +
-                "   assessmentKey DESC, \n" +
-                "   A.testPosition";
-
-        List<Map<String, Object>> rows = jdbcTemplate.queryForList(SQL, parameters);
-        Optional<Assessment> maybeAssessment = Optional.empty();
+        final String SQL = "SELECT \n" +
```
millerts, line 254: @jeffjohnson9046 , @wtritch , and I played around with a few queries around going from a segment to the overall assessment data, but the table is a mess.  Instead we get the assessment data from the segment then fetch the full assessment.

millerts: And note, we only have segment key at this point and this call originates from the existing item selection flow where segment key is all we have.

jeffjohnson9046 (approved):

service/src/main/java/tds/assessment/repositories/impl/AssessmentQueryRepositoryImpl.java
jtreuting: Makes sense to me ... i think 2 easy queries might even be faster in practice than 1 complex one

jtreuting (approved):
I'm not 100% sure I understand why the change is happening from parentItems to siblingItems, but assuming that needs to be done this looks good to me

millerts:
@jtreuting the main change was because I misread this [query](https://github.com/SmarterApp/TDS_TestDeliverySystemDataAccess/blob/master/tds-dll-mysql/src/main/java/tds/dll/mysql/ItemSelectionDLL.java#L2731).  During adaptive tests like CAT IRP Math 7 if the item can't be found in the current segment it tries to find it within the overall assessment.

This was causing errors in our flow because the parent assessment doesn't have items so "sibling" items were always empty.

PR #87, opened 2017-06-16
TDS-1048: Bugfix/get items by language

jeffjohnson9046:
[TDS-1048](https://jira.fairwaytech.com/browse/TDS-1048):  Refactored to handle items that have multiple languages.


client/src/main/java/tds/assessment/Item.java
```diff
@@ -222,21 +224,6 @@ public void setPrintable(boolean printable) {
     }
 
     /**
-     * Finds the language code of the item based on its item properties.
-     *
-     * @return The language code of the item if one exists in its properties - otherwise null
-     */
-    public String getLanguageCode() {
```
jeffjohnson9046, line 229: This property isn't used anywhere.  Which is good, because it doesn't represent that items can have multiple languages.  `TDS_ExamService` and `TDS_ExamResultsTransmitter` both depend on the `TDS_AssessmentService` client library, and both built fine after removing this property (made sure to point their `pom.xml` files to use the 3.1.1-SNAPSHOT version of this library to be sure).

millerts (approved):

PR #99, opened 2017-11-08
Remove assessment from configs/itembank tables

burzum619:
Quick note: 
For the moment, I'm assuming that test/assessment ids are unique and therefore we do not need to verify if other assessments (with different "assessment keys") share the same ID. If we get feedback from SBAC saying that they wish to maintain the possibility of sharing testids between assessments loaded into the system, I will make those updates in a separate PR.

The service method is wrapped in a `@Transactional` in case the delete call fails, in which case all changes would be rolled back by Spring. However, all the of the updates done in the `batchUpdate` call should be called sequentially within the transaction.

### DIFF
--- a/service/src/main/java/tds/assessment/repositories/impl/AssessmentCommandRepositoryImpl.java
+++ b/service/src/main/java/tds/assessment/repositories/impl/AssessmentCommandRepositoryImpl.java
@@ -19,7 +19,6 @@ import org.springframework.stereotype.Repository;
 
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collection;
 import java.util.List;
 import java.util.stream.Collectors;
 
@@ -70,6 +69,10 @@ public class AssessmentCommandRepositoryImpl implements AssessmentCommandReposit
     @Override
     public void removeAssessmentData(final String clientName, final Assessment assessment) {
         final List<String> SQL = new ArrayList<>();
+
+        if (!isClientAssessmentCombinationPresent(clientName, assessment.getAssessmentId())) {
+            return;
+        }
 
         // If the assessment is multi-segmented, remove segment data first
         if (assessment.isSegmented()) {
@@ -136,7 +139,7 @@ public class AssessmentCommandRepositoryImpl implements AssessmentCommandReposit
     }
 
     private List<String> generateConfigsDeleteQL(final String clientName, final String id) {
-        final List<String> SQL = Arrays.stream(CONFIGS_TABLE_NAMES)
+        return Arrays.stream(CONFIGS_TABLE_NAMES)
             .map(table -> {
                 switch (table) {
                     case "client_segmentproperties":
@@ -148,7 +151,11 @@ public class AssessmentCommandRepositoryImpl implements AssessmentCommandReposit
                 }
             })
             .collect(Collectors.toList());
+    }
 
-        return SQL;
+    private boolean isClientAssessmentCombinationPresent(final String clientName, final String assessmentId) {
+        // Execute this statement using a prepared statement to prevent SQL Injection attacks
+        return jdbcTemplate.queryForObject("SELECT COUNT(*) FROM configs.client_testproperties WHERE clientName = ? AND testid = ?",
+            new String[]{clientName, assessmentId}, Boolean.class);
     }
 }
--- a/service/src/main/java/tds/assessment/repositories/impl/AssessmentCommandRepositoryImpl.java
+++ b/service/src/main/java/tds/assessment/repositories/impl/AssessmentCommandRepositoryImpl.java
@@ -1,0 +1,140 @@
+/***************************************************************************************************
+ * Copyright 2017 Regents of the University of California. Licensed under the Educational
+ * Community License, Version 2.0 (the “license”); you may not use this file except in
+ * compliance with the License. You may obtain a copy of the license at
+ *
+ * https://opensource.org/licenses/ECL-2.0
+ *
+ * Unless required under applicable law or agreed to in writing, software distributed under the
+ * License is distributed in an “AS IS” BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for specific language governing permissions
+ * and limitations under the license.
+ **************************************************************************************************/
+
+package tds.assessment.repositories.impl;
+
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.jdbc.core.JdbcTemplate;
+import org.springframework.stereotype.Repository;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import tds.assessment.Assessment;
+import tds.assessment.repositories.AssessmentCommandRepository;
+
+@Repository
+public class AssessmentCommandRepositoryImpl implements AssessmentCommandRepository {
+    private final JdbcTemplate jdbcTemplate;
+    private static final String[] ITEMBANK_TABLE_NAMES = {
+        "aa_itemcl",
+        "tblsetofitemstrands",
+        "tblsetofitemstimuli",
+        "tblitemprops",
+        "setoftestgrades",
+        "testcohort",
+        "tblitemselectionparm",
+        "tbladminstrand",
+        "tblsetofadminitems",
+        "itemscoredimension",
+        "tbladminstimulus",
+        "testform",
+        "testformitem",
+        "affinitygroup",
+        "affinitygroupitem",
+        "tblsetofadminsubjects"
+    };
+
+    private static final String[] CONFIGS_TABLE_NAMES = {
+        "client_testmode",
+        "client_testformproperties",
+        "client_testwindow",
+        "client_testgrades",
+        "client_testeligibility",
+        "client_test_itemtypes",
+        "client_test_itemconstraint",
+        "client_testtooltype",
+        "client_segmentproperties",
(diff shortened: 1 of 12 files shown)


service/src/main/java/tds/assessment/repositories/impl/AssessmentCommandRepositoryImpl.java
```diff
@@ -0,0 +1,140 @@
...
+
+    private List<String> generateItembankDeleteSQL(final String assessmentKey) {
+        final List<String> SQL = Arrays.stream(ITEMBANK_TABLE_NAMES)
+            .map(table -> table.equals("tblsetofadminsubjects")
+                ? String.format("DELETE FROM itembank.%s WHERE _key = '%s';", table, assessmentKey)
+                : String.format("DELETE FROM itembank.%s WHERE _fk_adminsubject = '%s';", table, assessmentKey))
+            .collect(Collectors.toList());
+
+        // Delete items that are not being used by any assessment
+        SQL.add(
```
burzum619: The three SQL statements below are "special cases" where the data is not directly tied to an assessment within the target table. Instead, we will clear the "orphaned" data after the deletion of the assessment-specific data. The list is ordered, and I expect that JDBC/MySQL will honor this ordering an execute the batch statements in order.... at least that seems to be the case based on the manual testing I have performed locally.

service/src/main/java/tds/assessment/repositories/impl/AssessmentCommandRepositoryImpl.java
```diff
@@ -0,0 +1,140 @@
...
+                "       _fk_strand \n" +
+                "   FROM \n" +
+                "       tbladminstrand \n" +
+                ");"
+        );
+
+        return SQL;
+    }
+
+    private List<String> generateConfigsDeleteSQL(final String clientName, final String assessmentId) {
```
burzum619: I have these two private methods broken out based on the targeted database. I could easily combine the private methods if you guys think it might be cleaner.

service/src/test/java/tds/assessment/repositories/impl/AssessmentCommandRepositoryImplIntegrationTests.java
```diff
@@ -0,0 +1,74 @@
...
+    @Autowired
+    private JdbcTemplate jdbcTemplate;
+
+    @Value("classpath:(SBAC_PT)SBAC-IRP-CAT-ELA-3-Summer-2015-2016.xml")
+    private Resource res;
+
+    @Before
+    public void setup() throws IOException {
+        // Loads the test package
+        jdbcTemplate.execute(String.format("CALL itembank.loader_main('%s')", Files.toString(res.getFile(), Charsets.UTF_8)));
```
burzum619: I started to write some INSERT sql statements manually with test seed data, but pretty sure this ended up saving me at least half a day of some extremely tedious work.

service/src/test/java/tds/assessment/repositories/impl/AssessmentCommandRepositoryImplIntegrationTests.java
```diff
@@ -0,0 +1,74 @@
...
+    @Test
+    public void testRepository() {
+        Optional<Assessment> assessment = queryRepository.findAssessmentByKey(TEST_CLIENT_NAME, TEST_ASSESSMENT_KEY);
+        assertThat(assessment.isPresent()).isTrue();
+
+        commandRepository.removeAssessmentData(TEST_CLIENT_NAME, assessment.get());
+
+        Optional<Assessment> retAssessment = queryRepository.findAssessmentByKey(TEST_CLIENT_NAME, TEST_ASSESSMENT_KEY);
+        assertThat(retAssessment.isPresent()).isFalse();
+    }
```
trapnine: Deletes are dangerous. Add a test here and below (anywhere you're doing a delete) that also confirms that other innocent bystander assessments weren't removed by the delete. Also test with bad, partial, and wildcard input to the delete call to make sure it never blanket deletes.

service/src/test/java/tds/assessment/repositories/impl/AssessmentCommandRepositoryImplIntegrationTests.java
millerts: If this test fails for any reason is this cleaned up?  Will the loader tables have data in them?

trapnine (changes requested):
Need some side effect and negative testing as mentioned inline.

service/src/test/java/tds/assessment/repositories/impl/AssessmentCommandRepositoryImplIntegrationTests.java
millerts: Actually, how does this test run correctly?  For example, flyway doesn't create these stored procedures.  So how will they get on the dbs?  Or are we saying that everyone should make sure to run the db insert scripts on local?

If so, make sure to run that on the TeamCity database before merging this PR

service/src/main/java/tds/assessment/web/endpoints/AssessmentController.java
```diff
@@ -46,13 +48,18 @@ public AssessmentController(final AssessmentService service) {
...
             .orElseThrow(() -> new NotFoundException("Could not find set of admin subject for %s", key));
 
         return ResponseEntity.ok(assessment);
     }
 
-
+    @DeleteMapping(value = "/{clientName}/assessments/{key}", produces = MediaType.APPLICATION_JSON_VALUE)
+    ResponseEntity<?> removeAssessment(@PathVariable final String clientName,
+                                       @PathVariable final String key) throws NotFoundException {
+        service.removeAssessment(clientName, key);
```
millerts, line 62: What happens if you try to call this endpoint with:

```
/SBAC_PT/assessments/(SELECT _key FROM itembank.tblsetofadminsubjects limit 1)
```

Not sure what the encoding would be or how Spring would handle that.  I'm only asking because we're doing straight String building for SQL queries which can lead to SQL injection attacks.  Maybe this isn't a concern since this is contained within kubernetes and never exposed.

service/src/main/java/tds/assessment/repositories/impl/AssessmentCommandRepositoryImpl.java
```diff
@@ -0,0 +1,140 @@
...
+        final List<String> SQL = generateItembankDeleteSQL(assessment.getKey());
+        SQL.addAll(generateConfigsDeleteSQL(clientName, assessment.getAssessmentId()));
+
+        jdbcTemplate.batchUpdate(SQL.toArray(new String[SQL.size()]));
+    }
+
+    private List<String> generateItembankDeleteSQL(final String assessmentKey) {
+        final List<String> SQL = Arrays.stream(ITEMBANK_TABLE_NAMES)
+            .map(table -> table.equals("tblsetofadminsubjects")
+                ? String.format("DELETE FROM itembank.%s WHERE _key = '%s';", table, assessmentKey)
```
millerts: Is there not a way to use one of the other batch update methods that allow one to use prepared statements rather than building these queries up by hand?

burzum619: There is not unfortunately. The `NamedParameterJdbcTemplate` does not provide a method that allows multiple separate statements with `batchUpdate()`. @malaffoon Attempted to find a way around this and ran into a number of issues.

service/src/main/java/tds/assessment/web/endpoints/AssessmentController.java
burzum619: I did consider that. Not sure if Spring would process it, but even if it did we are doing a check at the service layer to see if the the assessment exists - so at worst it would return a 404.

service/src/test/java/tds/assessment/repositories/impl/AssessmentCommandRepositoryImplIntegrationTests.java
burzum619: If the test fails, things are cleaned up by Spring automatically because of the class-level `@Transactional` tag. But you do bring up a good point in your second comment. The `loader_main()` and other loader stored procedures may need to be added. This worked for me locally because I have all the stored procedures/tables/schemas created in my local mysql db.

service/src/test/java/tds/assessment/repositories/impl/AssessmentCommandRepositoryImplIntegrationTests.java
burzum619: I'll add additional test coverage to the integration tests. Thanks for the feedback.

service/src/test/java/tds/assessment/repositories/impl/AssessmentCommandRepositoryImplIntegrationTests.java
millerts: Ah, I missed the class level tag.  To the second point, what if we added the SQL dump to the flyway migration files?  This would include the table structure, stored procedures, and functions.  We could do that in every project if we wanted.  

That would give us two things:
1. It wouldn't force someone to manually run the schema themselves locally and could just use flyway
2. It would set this project up for the import work as whoever does that will need to do to the work of adding tests for import which can't use the stored procedure.

Maybe this isn't the time to do that and the person doing the insertion work can add that all in.  Whenever the create new procedure work is done I would expect this test to be modified to not use the loader_main procedures.  I think those procedures should be removed once this amendment is complete.

PR #100, opened 2017-11-14
Adding forcecomplete to the asssessment object to be used in force su…

millerts:
…bmission

### DIFF
--- a/service/src/main/java/tds/assessment/repositories/impl/AssessmentMapper.java
+++ b/service/src/main/java/tds/assessment/repositories/impl/AssessmentMapper.java
@@ -1,4 +1,4 @@
-/***************************************************************************************************
+/* **************************************************************************************************
  * Copyright 2017 Regents of the University of California. Licensed under the Educational
  * Community License, Version 2.0 (the “license”); you may not use this file except in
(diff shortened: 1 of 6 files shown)


service/src/main/java/tds/assessment/repositories/impl/AssessmentMapper.java
```diff
@@ -1,4 +1,4 @@
-/***************************************************************************************************
+/* **************************************************************************************************
```
millerts, line 1: This shows as a warning in IntelliJ since this uses JavaDoc style but isn't a JavaDoc.  I can revert this and change my settings to ignore this if we don't want this change.

service/src/main/java/tds/assessment/repositories/impl/AssessmentMapper.java
```diff
@@ -211,10 +212,6 @@ int getInt(String key) {
             return rowData.get(key) != null ? (int) rowData.get(key) : 0;
         }
 
-        long getLong(String key) {
```
millerts, line 214: Remove unused method.

trapnine (approved):

PR #102, opened 2018-02-02
API to load new TestPackage into TDS - Itembank

burzum619:
Added repositories, services, data model objects (mapped to each table), and a controller in order to add the capability to load the test package via HTTP. 

This PR is only for the `itembank` data - a separate story/PR will include the changes for the `configs` database. With that story complete ,the feature will be complete and we should be able to load and take assessments in TDS via this new endpoint.

A few notes:
- I did not add integration tests to the item bank repositories as they are simply spring-data-bound interfaces with no custom logic - these repositories are only used to simplify the inserts into each itembank table
- For each data object/table, I did not add javadocs to each properties' getter, as I'm not certain what many of the generic properties actually represent (and there are simply far too many in the entire item bank database). I did try to give a general idea of what each row represents in the class-level javadoc
- I followed my flow of the itembank_main() stored procedure (the `AssessmentLoaderService.loadTestPackage()` method being the entry point) but tried to break out each separate "procedure" into some service groupings that made sense.


service/src/main/java/tds/assessment/model/ItemMetadataWrapper.java
```diff
@@ -0,0 +1,74 @@
...
+import java.util.List;
+
+import tds.testpackage.model.Grade;
+import tds.testpackage.model.Item;
+
+/**
+ * A wrapper/helper class to enscapsulate various important pieces of metadata along with an item
+ */
+public class ItemMetadataWrapper {
+    private Item item;
```
mtrupkin: Fields may be final

service/src/main/java/tds/assessment/model/BlueprintElementTypes.java
```diff
@@ -0,0 +1,34 @@
...
+ * compliance with the License. You may obtain a copy of the license at
+ *
+ * https://opensource.org/licenses/ECL-2.0
+ *
+ * Unless required under applicable law or agreed to in writing, software distributed under the
+ * License is distributed in an “AS IS” BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for specific language governing permissions
+ * and limitations under the license.
+ **************************************************************************************************/
+
```
mtrupkin, line 13: One license per project is approved to be sufficient

mtrupkin (approved):
Looks simple for the most part.  It is hard to know if a different approach to implementing AssessmentLoaderService would simplify the logic.
Good job.

jeffjohnson9046 (approved):
The only quibble I have is that there are a couple of identity classes still named as "Identifier" (`TblAdminItemIdentifier` and `TblAdminStimulusIdentifier`).  It'd be nice to get those renamed to `TblAdminItemIdentity` and `TblAdminStimulusIdentity` for consistency, but I don't think that should hold up getting this PR merged in.

trapnine (approved):
LGTM. Look at all that boilerplate!

burzum619:
Closing out this PR without merging - a new PR will be opened later today that will include both the Itembank and Configs loading. All comments in this PR will be addressed in that PR

PR #103, opened 2018-02-13
Feature/assessment load - Loading assessment (itembank + configs dbs)

burzum619:
This PR is a combination of all itembank and configs changes, and contains some requested updates based on feedback for https://github.com/SmarterApp/TDS_AssessmentService/pull/102.

Many of the changes in this PR are boiler plate, but the more important parts of the code to focus on reviewing are any of the `*ConfigService*` classes that contain business logic for transforming and inserting data into the `configs` database. 

- There was some refactoring done to the `AssessmentLoaderServiceImpl` and `AssessmentItembankServiceImpl` classes - these had to be refactored so that `AssessmentItembankServiceImpl` and `AssessmentConfigsServiceImpl` were wrapped in individual `@Transactional` tags (with the rollback/cleanup managed by `AssessmentLoaderServiceImpl`). This was necessary because transactional writes are not supported with Hibernate to different schemas (without XA transactions). 

- There are some TODOs present in `AssessmentToolConfigServiceImpl.java` to uncomment some code. Some of the tool defaults will be moved to`TDS_SupportTool` and the `TestPackage` deserialization code will be updated to use this default values map to resolve these configuration flags. Once that code is in place, we should be able to uncomment the setters in this class.


service/src/main/java/tds/assessment/services/impl/AssessmentConfigSeedDataLoaderServiceImpl.java
```diff
@@ -0,0 +1,152 @@
...
+                        return segment.segmentForms().stream()
+                            .flatMap(form -> form.itemGroups().stream()
+                                .flatMap(itemGroup -> itemGroup.items().stream()
+                                    .flatMap(item -> item.getPresentations().stream()
+                                        //TODO: do we need a presentation @label attribute?
+                                        .map(language -> new Language(testPackage.getPublisher(), language.getCode(), language.label()))
+                                    )
+                                )
+                            );
+                    } else if (segment.getAlgorithmType().contains("adaptive")) {
```
jeffjohnson9046, line 107: Would it be possible to use `.equals(Algorithm.ADAPTIVE_2)` here instead of `.contains("adaptive")`?  Do we have adaptive selection algorithms that are not "adaptive2"?

service/src/main/java/tds/assessment/services/impl/AssessmentItemBankGenericDataLoaderServiceImpl.java
```diff
@@ -0,0 +1,200 @@
...
+        List<TblStimulus> stimuli = testPackage.getAssessments().stream().flatMap(
+            assessment -> assessment.getSegments().stream().flatMap(segment -> {
+                if (segment.getAlgorithmType().equalsIgnoreCase(Algorithm.FIXED_FORM.getType())) {
+                    return segment.segmentForms().stream()
+                        .flatMap(form -> form.itemGroups().stream()
+                            .filter(itemGroup -> itemGroup.getStimulus().isPresent())
+                            .map(ItemGroup::getStimulus)
+                            .map(stimulus -> mapStimuliToTblStimuli(bankKey, version, stimulus)));
+
+                } else if (segment.getAlgorithmType().contains("adaptive")) {
```
jeffjohnson9046, line 86: Same question here - are there selection algorithms besides "adaptive2"

service/src/main/java/tds/assessment/services/impl/AssessmentItemBankLoaderServiceImpl.java
```diff
@@ -0,0 +1,168 @@
...
+ * KIND, either express or implied. See the License for specific language governing permissions
+ * and limitations under the license.
+ **************************************************************************************************/
+
+package tds.assessment.services.impl;
+
+import com.google.common.collect.Lists;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
```
jeffjohnson9046, line 19: slf4j.Logger & LoggerFactory don't appear to be used anywhere

service/src/main/java/tds/assessment/services/impl/AssessmentSegmentLoaderServiceImpl.java
```diff
@@ -0,0 +1,334 @@
...
+                        .withIntercept(Float.parseFloat(itemSelectionProperties.getOrDefault("intercept", DEFAULT_INTERCEPT)))
+                        .withFieldTestStartPos(itemSelectionProperties.containsKey("ftstartpos")
+                            ? Integer.parseInt(itemSelectionProperties.get("ftstartpos"))
+                            : null)
+                        .withFieldTestEndPos(itemSelectionProperties.containsKey("ftendpos")
+                            ? Integer.parseInt(itemSelectionProperties.get("ftendpos"))
+                            : null)
+                        .withFieldTestMinItems(segmentBpElement.minFieldTestItems())
+                        .withFieldTestMaxItems(segmentBpElement.maxFieldTestItems())
+                        .withSelectionAlgorithm(segment.getAlgorithmType().equals("adaptive") ? Algorithm.ADAPTIVE_2.getType() : segment.getAlgorithmType())
```
jeffjohnson9046, line 177: Disregard my previous comments about using `.contains("adaptive")` - this answers my question(s) about using that as opposed to the `ADAPTIVE_2` enum value.

jeffjohnson9046 (approved):
A couple wee minor things and a question that I was answered myself.  As far as I'm concerned, none of my comments should prevent this PR from being merged into `develop`.  Nice work - and a lot of it!

service/src/main/java/tds/assessment/services/impl/AssessmentConfigSeedDataLoaderServiceImpl.java
burzum619: The legacy system accepts both "adaptive" and "adaptive2" as algorithm implementations. I think we may simply have to make it clear that "adaptive" and "fixedform" are currently the only acceptable values. We may have to programmatically map these to "adaptive2" in the TDS database during load though

service/src/main/java/tds/assessment/services/impl/AssessmentItemBankGenericDataLoaderServiceImpl.java
burzum619: This was basically done for backwards compatibility. I envision "adaptive" and "adaptive2" being synonymous for our purposes. "adaptive" is apparently an older version of the AIR adaptive algorithm that has been deprecated.

service/src/main/java/tds/assessment/services/impl/AssessmentSegmentLoaderServiceImpl.java
burzum619: Disregard my comments to your comments :)

PR #104, opened 2018-02-22
Additional changes to assessment service to support multiple assessme…

burzum619:
…nt deletes in one single API call

https://github.com/SmarterApp/TDS_SupportTool/pull/30 will block this from being merged.

This PR should also fix the serialization issue we are seeing currently in develop that prevents us from taking exams.


jeffjohnson9046 (approved):
Neat!

PR #106, opened 2018-03-15
TDS-1522: changed from 'blueprintweight' to match example XML

jeffjohnson9046:
I found this issue while working on TDS-1522 in TIS.  In every example test package XML file we have, the blueprint weight item selection property is named `bpweight` instead of `blueprintweight`.  In `tblSetOfAdminSubjects`, the column is indeed called `blueprintweight` (which is mapped correctly in the Hibernate config/annotations).  However, when attempting to fetch the blueprint weight value, we should (apprently) be looking for `bpweight` in the test package XML.

### DIFF
--- a/service/src/main/java/tds/assessment/services/impl/AssessmentSegmentLoaderServiceImpl.java
+++ b/service/src/main/java/tds/assessment/services/impl/AssessmentSegmentLoaderServiceImpl.java
@@ -175,7 +175,7 @@ public class AssessmentSegmentLoaderServiceImpl implements AssessmentSegmentLoad
                         .withFieldTestMinItems(segmentBpElement.minFieldTestItems())
                         .withFieldTestMaxItems(segmentBpElement.maxFieldTestItems())
                         .withSelectionAlgorithm(segment.getAlgorithmType().equals("adaptive") ? Algorithm.ADAPTIVE_2.getType() : segment.getAlgorithmType())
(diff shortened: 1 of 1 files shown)


service/src/main/java/tds/assessment/services/impl/AssessmentSegmentLoaderServiceImpl.java
```diff
@@ -175,7 +175,7 @@ public void loadAdminSubjects(final TestPackage testPackage, final String subjec
                         .withFieldTestMinItems(segmentBpElement.minFieldTestItems())
                         .withFieldTestMaxItems(segmentBpElement.maxFieldTestItems())
                         .withSelectionAlgorithm(segment.getAlgorithmType().equals("adaptive") ? Algorithm.ADAPTIVE_2.getType() : segment.getAlgorithmType())
-                        .withBlueprintWeight(Float.parseFloat(itemSelectionProperties.getOrDefault("blueprintweight", DEFAULT_BLUEPRINT_WEIGHT)))
+                        .withBlueprintWeight(Float.parseFloat(itemSelectionProperties.getOrDefault("bpweight", DEFAULT_BLUEPRINT_WEIGHT)))
```
jeffjohnson9046, line 178: I suppose another fix here would be to look for either `bpweight` or `blueprintweight`.  Unfortunately the XSD can't help us out here much; the `<itemSelection>` element can have any string for its `name` attribute.  But from what I've gathered, the convention for blueprint weight is `bpweight`.  Hence the change.

trapnine (approved):

PR #122, opened 2018-08-28
Ignoring duplicate dependencies that might be defined in a test package

burzum619:
Defining dependencies  in the test package that already exists causes weird UI issues in proctor where options are displayed multiple times for each tool. I am now filtering existing tools before I attempt to insert them.


service/src/main/java/tds/assessment/services/impl/AssessmentConfigSeedDataLoaderServiceImpl.java
```diff
@@ -92,10 +92,10 @@ public void loadSeedData(final TestPackage testPackage) {
...
 
-        // Load default tool dependency seed data
-        testPackage.getAssessments().forEach(assessment ->
-            jdbcTemplate.update("call configs.InsertToolDependencies(?, ?)", testPackage.getPublisher(), assessment.getId())
-        );
+//        // Load default tool dependency seed data
+//        testPackage.getAssessments().forEach(assessment ->
+//            jdbcTemplate.update("call configs.InsertToolDependencies(?, ?)", testPackage.getPublisher(), assessment.getId())
+//        );
     }
```
trapnine: Why is this commented?

trapnine (approved):
LGTM!

